PR #7, opened 2016-10-21
Change signature of methods overriding Future.then

natebosch:
The signature is:
`Future<T>.then (<S>((T) → dynamic, {onError: Function}) → Future<S>)`

Use `dynamic` rather than leave off the type to future proof against
`--no-implicit-dynamic`

cleanup: dartfmt

### DIFF
--- a/lib/src/delegate/future.dart
+++ b/lib/src/delegate/future.dart
@@ -25,13 +25,13 @@ class DelegatingFuture<T> implements Future<T> {
   Stream<T> asStream() => _future.asStream();
 
   Future<T> catchError(Function onError, {bool test(Object error)}) =>
-      _future.catchError(onError, test: test);
+    _future.catchError(onError, test: test);
 
   Future/*<S>*/ then/*<S>*/(dynamic onValue(T value), {Function onError}) =>
-      _future.then(onValue, onError: onError);
+    _future.then(onValue, onError: onError);
 
   Future<T> whenComplete(action()) => _future.whenComplete(action);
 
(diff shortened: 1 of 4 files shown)


natebosch:
This will help with https://github.com/dart-lang/bazel/issues/40

nex3:
Please remove dartfmt changes—this package doesn't use the formatter.

lib/src/delegate/future.dart
```diff
@@ -25,13 +25,13 @@ class DelegatingFuture<T> implements Future<T> {
   Stream<T> asStream() => _future.asStream();
 
   Future<T> catchError(Function onError, {bool test(Object error)}) =>
-    _future.catchError(onError, test: test);
+      _future.catchError(onError, test: test);
 
-  Future/*<S>*/ then/*<S>*/(/*=S*/ onValue(T value), {Function onError}) =>
-    _future.then(onValue, onError: onError);
+  Future/*<S>*/ then/*<S>*/(dynamic onValue(T value), {Function onError}) =>
```
nex3: What SDK version did this change in? We should set the minimum SDK constraint to that version.

natebosch: Looks like this particular bit changed in 1.19

nex3:
Can you publish this or should I?

natebosch:
Sorry meant to ping you about that right after I merged and then I got distracted.

If you grant me uploader permission I can publish and tag it, otherwise if you want to publish that works too.

PR #9, opened 2016-12-28
Add StreamQueue.startTransactions().

nex3:
This is important for advanced pull-based stream manipulation. It allows
users to express logic of the form "consume the next events if they
match this predicate".

### DIFF
--- a/lib/src/stream_queue.dart
+++ b/lib/src/stream_queue.dart
@@ -5,9 +5,12 @@
 import 'dart:async';
 import 'dart:collection';
 
+import 'package:collection/collection.dart';
+
 import "result.dart";
 import "subscription_stream.dart";
 import "stream_completer.dart";
+import "stream_splitter.dart";
 
 /// An asynchronous pull-based interface for accessing stream events.
 ///
@@ -86,7 +89,7 @@ abstract class StreamQueue<T> {
   bool _isClosed = false;
 
   /// Queue of events not used by a request yet.
-  final Queue<Result> _eventQueue = new Queue();
+  final QueueList<Result> _eventQueue = new QueueList();
 
   /// Queue of pending requests.
   ///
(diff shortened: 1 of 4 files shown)


lib/src/stream_queue.dart
```diff
@@ -407,6 +458,84 @@ class _StreamQueue<T> extends StreamQueue<T> {
...
+
+  /// Updates the parent [StreamQueue] to match this transaction's state.
+  ///
+  /// Further requests on this queue will be completed as though the underlying
+  /// stream had been closed, as though [cancel] were called with `immediate:
+  /// true`. Other transactions in the same batch are automatically rejected.
+  ///
+  /// Throws a [StateError] if [accept] or [reject] has already been called, or
+  /// if there are pending requests on this queue.
+  void accept() {
```
nex3: > (Or is this prevented because the `startTransaction` request would make the request-queue non-empty?)

Yep. An open transaction counts as a pending request.

lib/src/stream_queue.dart
```diff
@@ -407,6 +458,84 @@ class _StreamQueue<T> extends StreamQueue<T> {
...
+    _assertActive();
+
+    if (_requestQueue.isNotEmpty) {
+      throw new StateError(
+          "A transaction with pending requests can't be accepted.");
```
nex3: I went with the "just throw" solution because I didn't think it was really clear which of the several possible behaviors was the most obvious or most useful, and I couldn't come up with any really compelling use-cases that weren't supportable in other ways. If we throw for now it leaves us open to choose a different behavior in the future if it becomes clear that one is correct.

lib/src/stream_queue.dart
```diff
@@ -644,3 +771,91 @@ class _HasNextRequest<T> implements _EventRequest<T> {
...
+    for (var i = 0; i < count; i++) {
+      transactions[i] = new TransactionStreamQueue._(splitter.split(), this);
+    }
+    splitter.close();
+  }
+
+  bool update(QueueList<Result<T>> events, bool isDone) {
+    while (_eventsSent < events.length) {
+      var result = events[_eventsSent++];
+      if (result.isValue) {
```
nex3: Done.

lrhn:
LGTM!

PR #10, opened 2016-12-29
Add a subscriptionTransformer() function.

nex3:


lib/src/stream_subscription_transformer.dart
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'async_memoizer.dart';
+import 'delegate/stream_subscription.dart';
+
+typedef Future _CancelHandler<T>(StreamSubscription<T> inner);
```
nex3: Done.

lib/src/stream_subscription_transformer.dart
```diff
@@ -0,0 +1,71 @@
...
+
+/// Creates a [StreamTransformer] that modifies the behavior of subscriptions to
+/// a stream.
+///
+/// When [StreamSubscription.cancel], [StreamSubscription.pause], or
+/// [StreamSubscription.resume] is called, the corresponding handler is invoked.
+/// By default, handlers just forward to the underlying subscription.
+///
+/// This guarantees that once the transformed [StreamSubscription] has been
+/// canceled, no further callbacks will be invoked. The [handlePause] and
```
nex3: > The "This" here sounds like it refers to the entire previous paragraph, which doesn't guarantee that, e.g., if your cancel handler doesn't forward the cancel. If it only refers to the default, then it's not really interesting.
> 
> Should it say something like: 
> > It is guaranteed that ... 
> 
> ?
> (Probably not, since the code doesn't guarantee that).

Changed to just "Guarantees", which more clearly indicates that the function call itself provides the guarantee.

> So, it should be documented *very* heavily that user provided handlers MUST ensure that the inner stream doesn't provide any events while the outer subscription is paused or after it is cancelled.
> If they fail to do that, anything can happen. Failing the Stream contract can make code using `async for` completely unpredictable.

Done.

> Perhaps add a test after `handleCancel`/`handlePause` calls that `_inner` is paused or cancelled. (You can test `isPaused`, but to detect the cancel, you might need to wrap inner too, so that might not be worth it).
> Otherwise you can make sure that events happening while paused/cancelled cause async errors instead of being delivered.

I'd rather not add an additional layer of wrapping necessary to verify this.

lib/src/stream_subscription_transformer.dart
```diff
@@ -0,0 +1,71 @@
...
+/// already or not.
+StreamTransformer/*<T, T>*/ subscriptionTransformer/*<T>*/(
+    {Future handleCancel(StreamSubscription/*<T>*/ inner),
+    void handlePause(StreamSubscription/*<T>*/ inner),
+    void handleResume(StreamSubscription/*<T>*/ inner)}) {
+  return new StreamTransformer((stream, cancelOnError) {
+    return new _TransformedSubscription(
+        stream.listen(null, cancelOnError: cancelOnError),
+        handleCancel ?? (inner) => inner.cancel(),
+        handlePause ?? (inner) => inner.pause(),
```
nex3: Done.

lib/src/stream_subscription_transformer.dart
```diff
@@ -0,0 +1,71 @@
...
+  /// The callback to run when [pause] is called.
+  final _SimpleHandler<T> _handlePause;
+
+  /// The callback to run when [resume] is called.
+  final _SimpleHandler<T> _handleResume;
+
+  _TransformedSubscription(StreamSubscription<T> inner,
+          this._handleCancel, this._handlePause, this._handleResume)
+      : _inner = inner,
+        super(inner);
```
nex3: I don't think it's worth the extra complexity.

lib/src/stream_subscription_transformer.dart
```diff
@@ -18,10 +18,14 @@ typedef void _SimpleHandler<T>(StreamSubscription<T> inner);
 /// [StreamSubscription.resume] is called, the corresponding handler is invoked.
 /// By default, handlers just forward to the underlying subscription.
 ///
-/// This guarantees that once the transformed [StreamSubscription] has been
-/// canceled, no further callbacks will be invoked. The [handlePause] and
-/// [handleResume] are invoked regardless of whether the subscription is paused
-/// already or not.
+/// Guarantees that once the transformed [StreamSubscription] has been canceled,
+/// no further callbacks will be invoked. The [handlePause] and [handleResume]
```
lrhn: But, as I read it, that isn't guaranteed.

If the `onCancel` handler does nothing, you can get further event callbacks after calling `cancel` on the outer subscription. The only way to avoid that with the current code is for the `onCancel` to stop events from the inner subscription (either by cancel or pause). Removing the callbacks is one option, then at least you don't get callbacks, but since you can set them again, you also need to modify the `onData`, `onError` and `onDone` methods to not forward after a cancel.

For `pause` it's even worse because you can't remove the event handlers. 

So you, and this class, don't guarantee anything here. It is a requirement on the `handlePause` and `handleCancel` that they stop events.

lib/src/stream_subscription_transformer.dart
```diff
@@ -18,10 +18,14 @@ typedef void _SimpleHandler<T>(StreamSubscription<T> inner);
...
+/// are invoked regardless of whether the subscription is paused already or not.
+///
+/// In order to preserve [StreamSubscription] guarantees, **all callbacks must
+/// synchronously call the corresponding method** on the inner
```
lrhn: If that is the requirement, why don't you just do it for them, and let the handlers just be notifications?

I was assuming that the handlers were needed because you might want to *not* call the corresponding method, but do something else (e.g., direct events somewhere else during pause or pause instead of cancelling, so the subscription can be reused somewhere else - but then you really need to break the connection better, so you can't call through to the inner subscription after a cancel at all).

lib/src/stream_subscription_transformer.dart
nex3: When this says "no further callbacks will be invoked", it's referring to the callback arguments to this function—`handleCancel()`, `handlePause()`, and `handleResume()`. Those *are* guaranteed to be unused.

lib/src/stream_subscription_transformer.dart
nex3: I could imagine a user wanting to add some behavior after the `cancel()` future completes. I suppose we could make `handlePause`/`handleResume` notifications, but that would be oddly inconsistent.

lib/src/stream_subscription_transformer.dart
lrhn: OK, just make that extra clear (because it wasn't to me, and even I need to understand it :)

lib/src/stream_subscription_transformer.dart
lrhn: The one use-case I see where you don't want to recursively call cancel is that you want to reuse the subscription afterwards. That means that you might overwrite the callbacks or call `pause` *instead* of calling `cancel`.

For that to work safely, this wrapper should no longer be able to resume or set callbacks.
How about not making `_inner` final and setting it to `null` on cancel. Maybe also set the callbacks to `null` first. That should cut any connection to the wrapper (at least if you drop being a `DelegatingStreamSubscription` , and always use `_inner`).

nex3:
Ping!

lib/src/stream_subscription_transformer.dart
```diff
@@ -76,4 +99,7 @@ class _TransformedSubscription<T> extends DelegatingStreamSubscription<T> {
     if (_cancelMemoizer.hasRun) return;
     _handleResume(_inner);
   }
+
+  Future/*<E>*/ asFuture/*<E>*/([/*=E*/ futureValue]) =>
```
lrhn: You should be able to use real generics syntax now. It was released in 1.21.

lrhn:
LGTM!

lib/src/stream_subscription_transformer.dart
nex3: I'd like to hold off on this until 1.22 stable is released, to preserve flexibility in version solving.

Issue #8, opened 2016-12-27
FR: AsyncCache<T>, avoid duplicate async when fresh


matanlurey:
## Proposal

_As usual, happy to implement and send a PR_

A class, `SynchronizedFuture` (name debatable) that only calls a function once until completion:

```dart
abstract class SynchronizedFuture<T> implements Future<T> {
  factory AsyncSynchronizer(Future<T> call()) {
    // Use an expando and only create one per `call` function. See below
  }

  // Could also have another factory without a cache (similar to AsyncMemoizer use).
}
```

Example use:

```dart
class Example {
  // Actual implementation.
  Future<String> _doExpensiveCall() async {
    print('CALL MADE');
    return 'Hello';
  }

  // Visible API.
  Future<String> expensiveCall() => new SynchronizedFuture<String>(_doExpensiveCall);
}

main() async {
  var example = new Example();
  example.expensiveCall();
  example.expensiveCall();
  await new Future.value();
  // Output: 'CALL MADE', once.
  await example.expensiveCall();
  // Output: 'CALL MADE', again, since this is a _new_ call and none is in process
}
```

I'd also like to have a similar pattern for Streams, though that one would be a little more difficult since there isn't a true start/stop point - perhaps if at least one event has been returned on the stream it's considered "done" and new calls call the `call` function again for a new stream.

/cc @nex3

nex3:
I'm not sure I totally understand your example. You have three calls to `expensiveCall()`, but the comments indicate that "CALL MADE" is only printed twice. Does this mean the synchronizer would drop calls on the floor if a call was already in progress?

matanlurey:
What I mean is that it would just use the response of the (pending) future on calls being made before that future completes.

nex3:
That seems like it would produce a high risk of race conditions, since the difference between "use the value being computed" and "compute a new value from scratch" could just be a few microtasks—and presumably that distinction matters, or the user would just use `AsyncMemoizer` and always have the computed value available.

Maybe I'd understand better with a real-world example—what's the motivating use-case here?

matanlurey:
Sure, so let's say you have an "expensive operation*"

_*Read a large file from the disk, make a network request, spawn an isolate and do work_

To your users, you'd like to present a simplified API:

```dart
abstract class ExpensiveWorkService {
  Future<String> compute();
}
```

If a user calls `compute` multiple times though, you don't want to read/fetch/spawn multiple times.

[...]

PR #12, opened 2017-01-03
Add AsyncCache

matanlurey:
As discussed, and closes https://github.com/dart-lang/async/issues/8.

Will edit CHANGELOG and pubspec and squash before merge.

### DIFF
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -9,18 +9,18 @@ typedef DateTime _GetNow();
 /// Runs asynchronous functions and caches the result for a period of time.
 ///
 /// This class exists to cover the pattern of having potentially expensive code
-/// such as file I/O, network access, or isolate computation run fewer times in
-/// an application or process.
+/// such as file I/O, network access, or isolate computation that's unlikely to
+/// change quickly run.
 ///
 /// ## Example use with a `Future`:
 /// ```dart
-/// final _usersCache = new AsyncCache<List<String>>();
-///
 /// /// Runs the actual expensive code.
 /// Future<List<String>> _getOnlineUsers() => ...
 ///
+/// final _usersCache = new AsyncCache<List<String>>(const Duration(hours: 1));
+///
 /// /// Uses the cache if it exists, otherwise calls `_getOnlineUsers`.
-/// Future<List<String>> getOnlineUsers() => _usersCache.fetch(_getOnlineUsers);
+/// Future<List<String>> get onlineUsers => _usersCache.fetch(_getOnlineUsers);
 /// ```
 class AsyncCache<T> {
   static DateTime _defaultNow() => new DateTime.now();
@@ -28,10 +28,17 @@ class AsyncCache<T> {
   final Duration _duration;
   final _GetNow _getNow;
 
+  // Cached results of a previous `fetchStream` call.
   List<T> _cachedStreamValues;
+
+  // Non-null if a `fetchStream` call is currently in process, and completes
+  // once the stream is considered done.
   Future _fetchingStream;
 
+  // Cached results of a previous `fetch` call.
   Future<T> _cachedValueFuture;
+
+  // When a call to fetch or fetchStream should be automatically invalidated.
   DateTime _cachedExpiration;
 
   /// Creates a cache that invalidates contents after [duration] has passed.
@@ -43,8 +50,7 @@ class AsyncCache<T> {
   ///
   /// An ephemeral cache guarantees that a callback function will only be
   /// executed at most once concurrently. This is useful for requests which data
-  /// is updated frequently but the cost of making multiple requests without
-  /// previous completing yet is prohibitive or unwanted.
+  /// is updated frequently but stale data is acceptable.
   ///
   /// For _testing_, a custom `now` function may be provided.
   factory AsyncCache.ephemeral({DateTime now()}) =>
@@ -60,20 +66,29 @@ class AsyncCache<T> {
   /// If [callback] has been run recently enough, returns its previous return
   /// value. Otherwise, runs [callback] and returns its new return value.
   Future<T> fetch(Future<T> callback()) async {
+    if (_fetchingStream != null || _cachedStreamValues != null) {
+      throw new StateError('Previously used to cache via `fetchStream`');
+    }
     _invalidateWhenStale();
-    var result = await (_cachedValueFuture ??= callback());
+    if (_cachedValueFuture == null) {
+      _cachedValueFuture = callback();
+    }
+    var result = await _cachedValueFuture;
     _cachedExpiration ??= _getNow().add(_duration);
     return result;
   }
 
   /// Returns a cached stream or runs [callback] to compute a new stream.
   ///
-  /// If [callback] has been run recently enough, returns its previous return
-  /// value. Otherwise, runs [callback] and returns its new return value.
+  /// If [callback] has been run recently enough, returns a copy of its previous
+  /// return value. Otherwise, runs [callback] and returns its new return value.
   ///
   /// If a stream is currently being fetched, waits until the _done_ event and
   /// then returns the cached value.
   Stream<T> fetchStream(Stream<T> callback()) async* {
+    if (_cachedValueFuture != null) {
+      throw new StateError('Previously used to cache via `fetch`');
+    }
     _invalidateWhenStale();
     if (_fetchingStream != null) {
       await _fetchingStream;
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -1,0 +1,113 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+typedef DateTime _GetNow();
+
+/// Runs asynchronous functions and caches the result for a period of time.
+///
+/// This class exists to cover the pattern of having potentially expensive code
+/// such as file I/O, network access, or isolate computation run fewer times in
+/// an application or process.
+///
+/// ## Example use with a `Future`:
+/// ```dart
+/// final _usersCache = new AsyncCache<List<String>>();
+///
+/// /// Runs the actual expensive code.
+/// Future<List<String>> _getOnlineUsers() => ...
+///
+/// /// Uses the cache if it exists, otherwise calls `_getOnlineUsers`.
+/// Future<List<String>> getOnlineUsers() => _usersCache.fetch(_getOnlineUsers);
+/// ```
+class AsyncCache<T> {
+  static DateTime _defaultNow() => new DateTime.now();
+
+  final Duration _duration;
+  final _GetNow _getNow;
+
+  List<T> _cachedStreamValues;
+  Future _fetchingStream;
+
+  Future<T> _cachedValueFuture;
+  DateTime _cachedExpiration;
+
+  /// Creates a cache that invalidates contents after [duration] has passed.
+  ///
+  /// For _testing_, a custom `now` function may be provided.
+  factory AsyncCache(Duration duration, {DateTime now()}) = AsyncCache<T>._;
+
+  /// Creates a cache that invalidates after an in-flight request is complete.
+  ///
+  /// An ephemeral cache guarantees that a callback function will only be
+  /// executed at most once concurrently. This is useful for requests which data
+  /// is updated frequently but the cost of making multiple requests without
+  /// previous completing yet is prohibitive or unwanted.
+  ///
+  /// For _testing_, a custom `now` function may be provided.
+  factory AsyncCache.ephemeral({DateTime now()}) =>
+      new AsyncCache._(Duration.ZERO);
+
+  // Prevent inheritance. This allows using a redirecting factory constructor
+  // with different implementations in the future without a breaking API change.
+  AsyncCache._(this._duration, {DateTime now(): _defaultNow,})
+      : _getNow = now;
+
+  /// Returns a cached value or runs [callback] to compute a new one.
+  ///
+  /// If [callback] has been run recently enough, returns its previous return
+  /// value. Otherwise, runs [callback] and returns its new return value.
+  Future<T> fetch(Future<T> callback()) async {
+    _invalidateWhenStale();
+    var result = await (_cachedValueFuture ??= callback());
+    _cachedExpiration ??= _getNow().add(_duration);
+    return result;
+  }
+
+  /// Returns a cached stream or runs [callback] to compute a new stream.
+  ///
+  /// If [callback] has been run recently enough, returns its previous return
+  /// value. Otherwise, runs [callback] and returns its new return value.
+  ///
+  /// If a stream is currently being fetched, waits until the _done_ event and
+  /// then returns the cached value.
+  Stream<T> fetchStream(Stream<T> callback()) async* {
+    _invalidateWhenStale();
+    if (_fetchingStream != null) {
+      await _fetchingStream;
+    }
+    if (_cachedStreamValues != null) {
+      yield* new Stream<T>.fromIterable(_cachedStreamValues);
+      return;
+    }
+    var fetching = new Completer();
+    _fetchingStream = fetching.future;
+    var values = <T>[];
+    await for (var result in callback()) {
+      values.add(result);
+      yield result;
+    }
+    _cachedStreamValues = values;
+    _cachedExpiration ??= _getNow().add(_duration);
+    fetching.complete();
+  }
+
+  /// Removes any cached value, and returns `true` if a cached value existed.
+  Future<bool> invalidate() {
+    if (_cachedExpiration == null) {
+      return new Future<bool>.value(false);
+    }
+    _cachedExpiration = null;
+    _cachedValueFuture = null;
+    _cachedStreamValues = null;
+    return new Future<bool>.value(true);
+  }
+
+  void _invalidateWhenStale() {
+    if (_cachedExpiration != null && !_cachedExpiration.isAfter(_getNow())) {
+      invalidate();
+    }
+  }
+}
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -20,33 +20,42 @@ import 'package:async/async.dart';
 ///   // Actually fetch online users here.
 /// });
 /// ```
+///
+/// This class's timing can be mocked using [`fake_async`][fake_async].
+///
+/// [fake_async]: https://pub.dartlang.org/packages/fake_async
 class AsyncCache<T> {
-  // How long cached values stay fresh for.
+  /// How long cached values stay fresh.
   final Duration _duration;
 
-  // Cached results of a previous `fetchStream` call.
+  /// Cached results of a previous [fetchStream] call.
   StreamSplitter<T> _cachedStreamSplitter;
 
-  // Cached results of a previous `fetch` call.
+  /// Cached results of a previous [fetch] call.
   Future<T> _cachedValueFuture;
 
-  // Fires when the cache should be considered stale.
+  /// Fires when the cache should be considered stale.
   Timer _stale;
 
   /// Creates a cache that invalidates after an in-flight request is complete.
   ///
   /// An ephemeral cache guarantees that a callback function will only be
-  /// executed at most once concurrently. This is useful for requests which data
-  /// is updated frequently but stale data is acceptable.
-  factory AsyncCache.ephemeral({DateTime now()}) =>
+  /// executed at most once concurrently. This is useful for requests for which
+  /// data is updated frequently but stale data is acceptable.
+  factory AsyncCache.ephemeral() =>
       new AsyncCache(Duration.ZERO);
 
-  /// Creates a cache that invalidates contents after a duration has passed.
+  /// Creates a cache that invalidates its contents after [duration] has passed.
+  ///
+  /// The [duration] starts counting after the Future returned by [fetch]
+  /// completes, or after the Stream returned by [fetchStream] emits a done
+  /// event.
   AsyncCache(this._duration);
 
-  /// Returns a cached value or runs [callback] to compute a new one.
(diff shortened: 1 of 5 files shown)


lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+      : _getNow = now;
+
+  /// Returns a cached value or runs [callback] to compute a new one.
+  ///
+  /// If [callback] has been run recently enough, returns its previous return
+  /// value. Otherwise, runs [callback] and returns its new return value.
+  Future<T> fetch(Future<T> callback()) async {
+    _invalidateWhenStale();
+    var result = await (_cachedValueFuture ??= callback());
+    _cachedExpiration ??= _getNow().add(_duration);
```
nex3: I think this would be clearer if it were in an `if` block with the `_cachedValueFuture` assignment. It'd be more verbose, but right now it's hard to tell whether there are circumstances where one might be overwritten and the other might now.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+  // Prevent inheritance. This allows using a redirecting factory constructor
+  // with different implementations in the future without a breaking API change.
+  AsyncCache._(this._duration, {DateTime now(): _defaultNow,})
+      : _getNow = now;
+
+  /// Returns a cached value or runs [callback] to compute a new one.
+  ///
+  /// If [callback] has been run recently enough, returns its previous return
+  /// value. Otherwise, runs [callback] and returns its new return value.
+  Future<T> fetch(Future<T> callback()) async {
```
nex3: This should probably throw a `StateError` if a stream is cached, and vice versa.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+    _invalidateWhenStale();
+    var result = await (_cachedValueFuture ??= callback());
+    _cachedExpiration ??= _getNow().add(_duration);
+    return result;
+  }
+
+  /// Returns a cached stream or runs [callback] to compute a new stream.
+  ///
+  /// If [callback] has been run recently enough, returns its previous return
+  /// value. Otherwise, runs [callback] and returns its new return value.
```
nex3: "returns its previous return value" isn't accurate, since if it did that the stream would probably already have a listener. Maybe "returns a copy of its previous return value"?

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+    var fetching = new Completer();
+    _fetchingStream = fetching.future;
+    var values = <T>[];
+    await for (var result in callback()) {
+      values.add(result);
+      yield result;
+    }
+    _cachedStreamValues = values;
+    _cachedExpiration ??= _getNow().add(_duration);
+    fetching.complete();
```
nex3: I think you can use [`StreamSplitter`](https://www.dartdocs.org/documentation/async/1.11.3/async/StreamSplitter-class.html) to make this method simpler, and start replaying streams earlier as an added bonus. Just create a splitter on the first request, split a stream off on cache hits, and close the splitter when the cache expires.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+      await _fetchingStream;
+    }
+    if (_cachedStreamValues != null) {
+      yield* new Stream<T>.fromIterable(_cachedStreamValues);
+      return;
+    }
+    var fetching = new Completer();
+    _fetchingStream = fetching.future;
+    var values = <T>[];
+    await for (var result in callback()) {
```
nex3: Last I checked, `await for` was extremely slow on the VM :confused:. Also, it'll probably cause some unwanted behavior if the user cancels their subscription to the returned stream.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+      values.add(result);
+      yield result;
+    }
+    _cachedStreamValues = values;
+    _cachedExpiration ??= _getNow().add(_duration);
+    fetching.complete();
+  }
+
+  /// Removes any cached value, and returns `true` if a cached value existed.
+  Future<bool> invalidate() {
```
nex3: Why is this returning a Future?

test/async_cache_test.dart
```diff
@@ -0,0 +1,139 @@
...
+  DateTime time;
+
+  setUp(() {
+    // Create a cache that is fresh for an hour starting at 1/1/17 @ 12:30PM.
+    cache = new AsyncCache<String>(const Duration(hours: 1), now: () => time);
+    time = new DateTime(2017, 1, 1, 12, 30);
+  });
+
+  test('should fetch via a callback when no cache exists', () async {
+    var callback = expectAsync0/*<Future<String>>*/(() async => 'Expensive');
```
nex3: I'd make this callback inline below, which should eliminate the need for an explicit generic type.

test/async_cache_test.dart
```diff
@@ -0,0 +1,139 @@
...
+    expect(await cache.fetch(callback), 'Expensive');
+  });
+
+  test('should not fetch via callback when a cache exists', () async {
+    Future<String> doNotCall() async => 'Should not have been invoked...';
```
nex3: Make this inline, and use `expectAsync` to assert that it's actually not called (e.g. `expectAsync0(() {}, count: 0)`). Same for callbacks below that aren't intended to be called.

test/async_cache_test.dart
```diff
@@ -0,0 +1,139 @@
...
+    Stream<String> call() {
+      return new Stream<String>.fromIterable(['Called ${++timesCalled}']);
+    }
+    expect(await cache.fetchStream(call).toList(), ['Called 1']);
+    await cache.invalidate();
+    expect(await cache.fetchStream(call).toList(), ['Called 2']);
+    await cache.invalidate();
+    expect(await cache.fetchStream(call).toList(), ['Called 3']);
+  });
+}
```
nex3: Add tests that assert that canceling or pausing one stream doesn't affect others.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+  /// previous completing yet is prohibitive or unwanted.
+  ///
+  /// For _testing_, a custom `now` function may be provided.
+  factory AsyncCache.ephemeral({DateTime now()}) =>
+      new AsyncCache._(Duration.ZERO);
+
+  // Prevent inheritance. This allows using a redirecting factory constructor
+  // with different implementations in the future without a breaking API change.
+  AsyncCache._(this._duration, {DateTime now(): _defaultNow,})
+      : _getNow = now;
```
matanlurey: Unfortunately I've seen inheritance as very common internally if allowed :-/

nex3: It's still not the style externally to forbid it using private constructors.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+typedef DateTime _GetNow();
+
+/// Runs asynchronous functions and caches the result for a period of time.
+///
+/// This class exists to cover the pattern of having potentially expensive code
+/// such as file I/O, network access, or isolate computation run fewer times in
```
matanlurey: Done

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+/// an application or process.
+///
+/// ## Example use with a `Future`:
```
matanlurey: Where would a good place to put this be?

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+/// ```dart
+/// final _usersCache = new AsyncCache<List<String>>();
```
matanlurey: Done.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+///
+/// /// Runs the actual expensive code.
+/// Future<List<String>> _getOnlineUsers() => ...
+///
+/// /// Uses the cache if it exists, otherwise calls `_getOnlineUsers`.
+/// Future<List<String>> getOnlineUsers() => _usersCache.fetch(_getOnlineUsers);
```
matanlurey: Done.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+  static DateTime _defaultNow() => new DateTime.now();
+
+  final Duration _duration;
+  final _GetNow _getNow;
+
+  List<T> _cachedStreamValues;
+  Future _fetchingStream;
+
+  Future<T> _cachedValueFuture;
+  DateTime _cachedExpiration;
```
matanlurey: Done.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  /// Creates a cache that invalidates contents after [duration] has passed.
+  ///
+  /// For _testing_, a custom `now` function may be provided.
+  factory AsyncCache(Duration duration, {DateTime now()}) = AsyncCache<T>._;
```
matanlurey: Do we need to do this here? Certainly needed for testability and certainly we don't have a standard way of doing this yet.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  /// Creates a cache that invalidates after an in-flight request is complete.
+  ///
+  /// An ephemeral cache guarantees that a callback function will only be
+  /// executed at most once concurrently. This is useful for requests which data
+  /// is updated frequently but the cost of making multiple requests without
+  /// previous completing yet is prohibitive or unwanted.
```
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: What do you mean? I've seen plenty of packages do this including the core SDK.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
```diff
@@ -0,0 +1,113 @@
...
+  AsyncCache._(this._duration, {DateTime now(): _defaultNow,})
+      : _getNow = now;
+
+  /// Returns a cached value or runs [callback] to compute a new one.
+  ///
+  /// If [callback] has been run recently enough, returns its previous return
+  /// value. Otherwise, runs [callback] and returns its new return value.
+  Future<T> fetch(Future<T> callback()) async {
+    _invalidateWhenStale();
+    var result = await (_cachedValueFuture ??= callback());
```
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Is there a bug filed for this? Why do we have language features that are performance prohibitive to use?

lib/src/async_cache.dart
matanlurey: I could see a sub-type of `AsyncCache` where the value(s) are cached on something like `IndexedDb` and aren't accessible (or invalidtable) synchronously. WDUT?

test/async_cache_test.dart
matanlurey: Done.

test/async_cache_test.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done! Love it!

test/async_cache_test.dart
matanlurey: Done.

matanlurey:
Friendly ping to @nex3 :)

lib/src/async_cache.dart
nex3: The SDK has very different constraints for backwards-compatibility than packages do. As far as other packages go, I haven't seen any that do this kind of hack to seal a class—maybe some exist, but this package (and other core extension packages) isn't one of them.

Besides, I'm the maintainer and I'll deal with the consequences if people end up extending this and causing problems—no extra work for you 😊.

lib/src/async_cache.dart
```diff
@@ -43,70 +48,76 @@ class AsyncCache<T> {
...
+    if (_cachedStreamSplitter != null) {
+      throw new StateError('Previously used to cache via `fetchStream`');
+    }
     _invalidateWhenStale();
-    var result = await (_cachedValueFuture ??= callback());
+    if (_cachedValueFuture == null) {
+      _cachedValueFuture = callback();
+    }
+    var result = await _cachedValueFuture;
     _cachedExpiration ??= _getNow().add(_duration);
```
nex3: I was thinking this would look like:

```dart
if (_cachedValueFuture == null) {
  _cachedValueFuture = callback();
  await _cachedValueFuture;
  _cachedExpiration = _getNow().add(_duration);
}
```

so it's clear that `_cachedExpiration` is reset if and only if there wasn't a cached value.

lib/src/async_cache.dart
```diff
@@ -43,70 +48,76 @@ class AsyncCache<T> {
...
-    return new Future<bool>.value(true);
+    _cachedStreamSplitter = null;
+    return new Future<bool>.value(wasCached);
+  }
+
+  bool get _isInvalidated {
+    return _cachedExpiration != null && !_cachedExpiration.isAfter(_getNow());
   }
 
   void _invalidateWhenStale() {
```
nex3: It feels strange that we're lazily invalidating the cache on access, rather than eagerly doing it with a timer. I worry that we'll end up keeping large amounts of data in memory uselessly.

lib/src/async_cache.dart
```diff
@@ -43,70 +48,76 @@ class AsyncCache<T> {
...
     _cachedExpiration = null;
     _cachedValueFuture = null;
-    _cachedStreamValues = null;
-    return new Future<bool>.value(true);
+    _cachedStreamSplitter = null;
+    return new Future<bool>.value(wasCached);
+  }
+
+  bool get _isInvalidated {
+    return _cachedExpiration != null && !_cachedExpiration.isAfter(_getNow());
```
nex3: I think `_getNow().isAfter(_cachedExpiration)` would read a little better.

lib/src/async_cache.dart
```diff
@@ -43,70 +48,76 @@ class AsyncCache<T> {
...
-      await _fetchingStream;
+  Stream<T> fetchStream(Stream<T> callback()) {
+    if (_cachedValueFuture != null) {
+      throw new StateError('Previously used to cache via `fetch`');
     }
-    if (_cachedStreamValues != null) {
-      yield* new Stream<T>.fromIterable(_cachedStreamValues);
-      return;
+    if (_isInvalidated) {
+      _cachedStreamSplitter.close();
```
nex3: Consider moving this into `_invalidateWhenStale()` (as `_cachedStreamSplitter?.close()`) so that you can just call that here.

lib/src/async_cache.dart
```diff
@@ -43,70 +48,76 @@ class AsyncCache<T> {
...
-    _fetchingStream = fetching.future;
-    var values = <T>[];
-    await for (var result in callback()) {
-      values.add(result);
-      yield result;
+    if (_cachedStreamSplitter == null) {
+      _cachedStreamSplitter = new StreamSplitter<T>(callback());
+      _cachedStreamSplitter.split().last.then((_) {
+        _cachedExpiration = _getNow().add(_duration);
+      });
```
nex3: `.last` will throw if the stream emits an error, and then that error will be unhandled. You could use `.listen(null, onDone: ...)` to avoid that.

That said, eagerly listening here breaks some of the laziness benefits of streams and `StringSplitter`'s nice cancel/pause handling. If you transform the input stream, you can avoid that:

```dart
new StreamSplitter(callback().transform(
    new StreamTransformer.fromHandlers(handleDone: (sink) {
  _cachedExpiration = _getNow().add(_duration);
})))
```

lib/src/async_cache.dart
nex3: > Is there a bug filed for this?

Yeah, dart-lang/sdk#23645.

> Why do we have language features that are performance prohibitive to use?

I don't think the feature is inherently prohibitive; the implementations we have just aren't tuned. The reasons behind that are above my pay grade.

lib/src/async_cache.dart
nex3: Even in that case, you could make `invalidate()` synchronously set a flag that causes further calls to `fetch()` to wait until the internal invalidation is done before fetching new info.

test/async_cache_test.dart
```diff
@@ -13,17 +13,16 @@ void main() {
...
 
   test('should fetch via a callback when no cache exists', () async {
-    var callback = expectAsync0/*<Future<String>>*/(() async => 'Expensive');
-    expect(await cache.fetch(callback), 'Expensive');
+    expect(await cache.fetch(() async => 'Expensive'), 'Expensive');
   });
 
   test('should not fetch via callback when a cache exists', () async {
-    Future<String> doNotCall() async => 'Should not have been invoked...';
+    var doNotCall = expectAsync0/*<Future<String>>*/((){}, count: 0);
```
nex3: Nit: include these inline in the `cache.fetch()` calls. That'll avoid the need to name them, and the need to explicitly annotate.

test/async_cache_test.dart
```diff
@@ -136,4 +132,25 @@ void main() {
     await cache.invalidate();
     expect(await cache.fetchStream(call).toList(), ['Called 3']);
   });
+
+  test('should cancel a cached stream without effecting others', () async {
```
nex3: "affecting" :wink:

test/async_cache_test.dart
```diff
@@ -136,4 +132,25 @@ void main() {
...
+    Stream<String> call() {
+      return new Stream<String>.fromIterable(['1', '2', '3']);
```
nex3: This shouldn't need an explicit generic type.

test/async_cache_test.dart
```diff
@@ -136,4 +132,25 @@ void main() {
...
+    }
+    expect(cache.fetchStream(call).toList(), completion(['1', '2', '3']));
+    expect(await cache.fetchStream(call).first, '1');
```
nex3: Add a note that `.first` cancels the stream.

test/async_cache_test.dart
```diff
@@ -136,4 +132,25 @@ void main() {
...
+  test('should pause a cached stream without effecting others', () async {
+    Stream<String> call() {
+      return new Stream<String>.fromIterable(['1', '2', '3']);
+    }
+    StreamSubscription sub;
+    sub = cache.fetchStream(call).listen((event) {
+      if (event == '1') {
+        sub.pause();
+      }
+    });
```
nex3: Wrap this callback in `expectAsync` so we can be sure the stream is actually getting paused.

lib/src/async_cache.dart
matanlurey: I guess my question is should I gate landing this with a testable API on adding a dependency on pkg/clock, which looks like it hasn't been maintained since 2014? I could also remove `now` and say manually call `invalidate` until we figure out our strategy. WDUT?

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Changed to a synchronous `void` for now then.

lib/src/async_cache.dart
```diff
@@ -4,34 +4,39 @@
...
+
 typedef DateTime _GetNow();
 
 /// Runs asynchronous functions and caches the result for a period of time.
 ///
 /// This class exists to cover the pattern of having potentially expensive code
-/// such as file I/O, network access, or isolate computation run fewer times in
-/// an application or process.
+/// such as file I/O, network access, or isolate computation that's unlikely to
+/// change quickly run.
```
matanlurey: Done.

lib/src/async_cache.dart
```diff
@@ -4,34 +4,39 @@
...
+///
 /// /// Uses the cache if it exists, otherwise calls `_getOnlineUsers`.
-/// Future<List<String>> getOnlineUsers() => _usersCache.fetch(_getOnlineUsers);
+/// Future<List<String>> get onlineUsers => _usersCache.fetch(_getOnlineUsers);
 /// ```
 class AsyncCache<T> {
   static DateTime _defaultNow() => new DateTime.now();
 
   final Duration _duration;
   final _GetNow _getNow;
```
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: Done.

lib/src/async_cache.dart
matanlurey: I thought about this too, and I agree now. This also means you can use fake async to invalidate the cache time instead of injecting a clock. Done.

test/async_cache_test.dart
matanlurey: Done.

test/async_cache_test.dart
matanlurey: Affected change ;)

test/async_cache_test.dart
matanlurey: Done here and elsewhere.

test/async_cache_test.dart
matanlurey: Done.

nex3:
Thanks Matan!

PR #13, opened 2017-01-03
Additional strong-mode fixes (down to 0 after this PR)

matanlurey:
/cc @natebosch

### DIFF
--- a/lib/src/result.dart
+++ b/lib/src/result.dart
@@ -37,7 +37,7 @@ abstract class Result<T> {
   /// The result of the transformation is a sink that only forwards [Result]
   /// values and no error events.
   static const StreamSinkTransformer<Object, Result> captureSinkTransformer =
-      const StreamSinkTransformer.fromStreamTransformer(
+      const StreamSinkTransformer<Object, Result>.fromStreamTransformer(
           const CaptureStreamTransformer());
 
   /// A sink transformer that releases result events.
@@ -97,7 +97,7 @@ abstract class Result<T> {
   /// The returned stream will not have any error events.
   /// Errors from the source stream have been converted to [ErrorResult]s.
   static Stream<Result/*<T>*/> captureStream/*<T>*/(Stream/*<T>*/ source) =>
-      source.transform(new CaptureStreamTransformer/*<T>*/());
+      source.transform/*<Result<T>>*/(new CaptureStreamTransformer/*<T>*/());
 
   /// Release a stream of [result] values into a stream of the results.
   ///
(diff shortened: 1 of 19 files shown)


lib/src/result.dart
```diff
@@ -37,7 +37,7 @@ abstract class Result<T> {
   /// The result of the transformation is a sink that only forwards [Result]
   /// values and no error events.
   static const StreamSinkTransformer<Object, Result> captureSinkTransformer =
-      const StreamSinkTransformer.fromStreamTransformer(
```
nex3: @jmesserly Why isn't this inferred? Same with the other changes in this file and `stream_sink_transformer.dart`.

natebosch: I think because `const CaptureStreamTransformer()` is a `StreamTransformer<dynamic, Result>` and not a `StreamTransformer<Object, Result>`

I *think* we could fix this instead with `const CaptureStreamTransformer<Object>()`

jmesserly: without seeing the error msg, it's hard to say for sure. I don't think dynamic vs Object would itself be problematic because those are treated almost identically for the subtype relation. You could try `const CaptureStreamTransformer<Object>()` and see if that fixes the problem.

Perhaps also worth trying without the `const`. `const` goes down pretty different code paths in Analyzer.

jmesserly: (the fact that the type annotation is required looks like some sort of bug to me, but of what sort or whether we have it filed already, I'm not certain)

matanlurey: Yes there was a warning. I'm happy to file a tracking bug.

jmesserly: if you could, thanks! ideally w/ enough info to reproduce & things you tried to make the warning go away.

test/stream_sink_transformer_test.dart
```diff
@@ -139,16 +139,16 @@ void main() {
     test("transforms error events", () {
       var transformer = new StreamSinkTransformer.fromHandlers(
           handleError: (i, stackTrace, sink) {
-            sink.addError(i * 2, stackTrace);
+            sink.addError((i as num) * 2, stackTrace);
```
natebosch: does it work to do instead:

```
handleError: (num i, stackTrace, sink) {
```

matanlurey: Not valid in checked-mode without covariant override

lib/src/result.dart
matanlurey: So this went away, somehow. I have no idea why. Reverted changes.

PR #19, opened 2017-01-20
Add StreamQueue.eventsDispatched.

nex3:
This is useful for advanced transaction operations, such as "one of the
following consumers accepts the queue". We want to reliably accept the
same consumer every time, and the best way to do that is to accept the
one that consumed the most events.

### DIFF
--- a/lib/src/stream_queue.dart
+++ b/lib/src/stream_queue.dart
@@ -89,6 +89,15 @@ abstract class StreamQueue<T> {
   /// Closing operations are [cancel] and [rest].
   bool _isClosed = false;
 
+  /// The number of events dispatched by this queue.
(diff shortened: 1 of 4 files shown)


lib/src/stream_queue.dart
```diff
@@ -628,25 +636,6 @@ class StreamQueueTransaction<T> {
   }
 }
 
-/// A [StreamQueue] that belongs to a [StreamQueueTransaction].
-class _TransactionStreamQueue<T> extends _StreamQueue<T> {
```
lrhn: Losing the special-casing. Wohoo!

lrhn:
LGTM!

PR #22, opened 2017-02-01
Bug a bug in StreamQueueTransaction.reject().

nex3:
This would throw a StateError if StreamQueue.rest had been called on
one of the transaction's child queues, because that queue didn't expect
to be canceled later on.

### DIFF
--- a/test/stream_queue_test.dart
+++ b/test/stream_queue_test.dart
@@ -863,6 +863,30 @@ main() {
         transaction.reject();
       });
 
+      // Regression test.
+      test("pending child rest requests emit no more events", () async {
+        var controller = new StreamController();
+        var events = new StreamQueue(controller.stream);
+        var transaction = events.startTransaction();
+        var queue = transaction.newQueue();
+
+        // This should emit no more events after the transaction is rejected.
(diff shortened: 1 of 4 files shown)


test/stream_queue_test.dart
```diff
@@ -863,6 +863,30 @@ main() {
...
+        transaction.reject();
+        await flushMicrotasks();
+
+        // These shouldn't affect the result of `queue.rest.toList()`.
+        controller.add(4);
+        controller.add(5);
+      });
+
       test("child requests' cancel() may still be called explicitly", () async {
         transaction.reject();
```
lrhn: Unrelated comment: It's probably too late, but could we consider renaming `reject` to `cancel`? It's just so much more common, and it was the first name I looked for when I tried using the class.

lrhn:
LGTM

test/stream_queue_test.dart
nex3: I think it's probably too late. I'm not sure I like `cancel` anyway—in other contexts it usually refers to finalization and releasing of resources, whereas for a transaction it's much more part of the normal flow of usage.

PR #27, opened 2017-02-14
Use FutureOr.

### LINKED ISSUE #26: Implementations of Future.then should be updated to use FutureOr
Classes that implement `Future` should change the signature ot `.then` to be

```dart
then(FutureOr<S> onValue(T x), {Function onError})
```

to ensure that inference will continue to work for calls to `.then` and to get better error checking.

nex3:
Closes #26

### DIFF
--- a/lib/src/delegate/future.dart
+++ b/lib/src/delegate/future.dart
@@ -27,7 +27,7 @@ class DelegatingFuture<T> implements Future<T> {
   Future<T> catchError(Function onError, {bool test(Object error)}) =>
     _future.catchError(onError, test: test);
 
-  Future/*<S>*/ then/*<S>*/(dynamic onValue(T value), {Function onError}) =>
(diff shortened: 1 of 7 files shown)


lib/src/delegate/future.dart
```diff
@@ -27,7 +27,7 @@ class DelegatingFuture<T> implements Future<T> {
   Future<T> catchError(Function onError, {bool test(Object error)}) =>
     _future.catchError(onError, test: test);
 
-  Future/*<S>*/ then/*<S>*/(dynamic onValue(T value), {Function onError}) =>
+  Future/*<S>*/ then/*<S>*/(FutureOr/*<S>*/ onValue(T value), {Function onError}) =>
```
kevmoo: Might as well use full generic method syntax, too – right?

nex3: Lasse has a separate CL out for that.

lrhn: I have landed that now, so you can update the one on `FutureOr`.

lrhn:
LGTM

PR #29, opened 2017-04-28
Add stream_transformers library and buffer utility

natebosch:
- Add stream_transformers library marked `@experimental` so we can
  iterate a bit on the APIs.
- Add `_Buffer` implementation which can collect values from a source
  stream and only emit them when a 'trigger' stream fires.
- Add top level method `buffer` which instantiates the `_Buffer` class.
  The stream transformers in this library will all be exposed as a top
  level method.

### DIFF
--- a/lib/src/stream_transformers/buffer.dart
+++ b/lib/src/stream_transformers/buffer.dart
@@ -1,0 +1,119 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+/// Creates a [StreamTransformer] which collects values and emits when it sees a
+/// value on [trigger].
+///
+/// If there are no pending values when [trigger] emits, the next value on the
+/// source Stream will immediately flow through. Otherwise, the pending values
+/// are released when [trigger] emits.
+StreamTransformer<T, List<T>> buffer<T>(Stream trigger) =>
+    new _Buffer(trigger, _collectToList);
+
+List<T> _collectToList<T>(T element, List<T> soFar) {
+  soFar ??= <T>[];
+  soFar.add(element);
+  return soFar;
+}
+
+/// A strategy for aggregating values.
+typedef R _Collect<T, R>(T element, R soFar);
+
+/// A StreamTransformer which aggregates values and emits when it sees a value
+/// on [_trigger].
+///
+/// If there are no pending values when [_trigger] emits the first value on the
+/// source Stream will immediately flow through. Otherwise, the pending values
+/// and released when [_trigger] emits.
+class _Buffer<T, R> implements StreamTransformer<T, R> {
+  final Stream _trigger;
+  final _Collect _collect;
+
+  _Buffer(this._trigger, this._collect);
+
+  @override
+  Stream<R> bind(Stream<T> values) {
+    StreamController<R> controller;
+    if (values.isBroadcast) {
+      controller = new StreamController<R>.broadcast();
+    } else {
+      controller = new StreamController<R>();
+    }
+
+    R currentResults;
+    bool waitingForValue = false;
+    StreamSubscription valuesSub;
+    StreamSubscription triggerSub;
+
+    cancelValues() {
+      var sub = valuesSub;
+      valuesSub = null;
+      return sub?.cancel() ?? new Future.value();
+    }
+
+    cancelTrigger() {
+      var sub = triggerSub;
+      triggerSub = null;
+      return sub?.cancel() ?? new Future.value();
+    }
+
+    closeController() {
+      var ctl = controller;
+      controller = null;
+      return ctl?.close() ?? new Future.value();
+    }
+
+    controller.onListen = () {
+      emit() {
+        controller.add(currentResults);
+        currentResults = null;
+      }
+
+      valuesSub = values.listen(
+          (T value) {
+            currentResults = _collect(value, currentResults);
+            if (waitingForValue) {
+              emit();
+              waitingForValue = false;
+            }
+          },
+          onError: controller.addError,
+          onDone: () {
+            valuesSub = null;
+          });
+      triggerSub = _trigger.listen(
+          (_) {
+            if (currentResults == null) {
+              waitingForValue = true;
+              return;
+            }
+            emit();
+            if (valuesSub == null) {
+              controller.close();
+              cancelTrigger();
+            }
+          },
+          onError: controller.addError,
+          onDone: () async {
+            cancelValues();
+            closeController();
+          });
+    };
+
+    // Forward methods from listener
+    if (!values.isBroadcast) {
+      controller.onPause = () {
+        valuesSub?.pause();
+        triggerSub?.pause();
+      };
+      controller.onResume = () {
+        valuesSub?.resume();
+        triggerSub?.resume();
+      };
+    }
+    controller.onCancel = () => Future.wait([cancelValues(), cancelTrigger()]);
+    return controller.stream;
+  }
+}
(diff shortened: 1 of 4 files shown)


natebosch:
@matanlurey - after some discussion I think this is going to be the right place to land some of those common utilities we've discussed in the past.

I'm happy to iterate a bit on the API. I implemented `_Buffer` to take the aggregation approach as an argument since it was trivial to do so, but `buffer` only works with a default argument of collecting to a list since that matches the semantics I think most folks would expect.

I was hoping to be able to do `StreamTransformer<T,R> buffer<T,R>(Stream<T> values, [R collect(T value, R soFar) = _collectToList])` but the type system does not allow me to express "limit `R` to `List<T>` if no `collect` is passed, otherwise use the type argument from `collect`"

lib/src/stream_transformers/buffer.dart
```diff
@@ -0,0 +1,133 @@
...
+      }
+    }
+
+    triggerDone() {
+      cancelValues();
+      closeController();
+    }
+
+    controller.onListen = () {
+      valuesSub = values.listen(onValue,
```
jakemac53: I don't know too much about broadcast streams, but should this bail if `valuesSub != null`?

lib/src/stream_transformers/buffer.dart
```diff
@@ -0,0 +1,133 @@
...
+          onError: controller.addError, onDone: valuesDone);
+      triggerSub = _trigger.listen(onTrigger,
+          onError: controller.addError, onDone: triggerDone);
+    };
+
+    // Forward methods from listener
+    if (!values.isBroadcast) {
```
jakemac53: Should we be trying to track the number of current listeners for broadcast streams and pause if it goes to zero?

test/stream_transformers/buffer_test.dart
```diff
@@ -0,0 +1,138 @@
+import 'dart:async';
```
jakemac53: nit: copyright

test/stream_transformers/buffer_test.dart
```diff
@@ -0,0 +1,138 @@
...
+    var values = new StreamController();
+    trigger.add(null);
+    var hasEmitted = false;
+    values.stream.transform(buffer(trigger.stream)).listen((_) {
+      hasEmitted = true;
+    });
+    values.add(1);
+    await new Future(() {});
+    expect(hasEmitted, true);
+  });
```
jakemac53: I would add one more test that goes through more than one iteration of collecting values and getting triggers.

jakemac53: Oh also a broadcast stream test would be good

lib/src/stream_transformers/buffer.dart
```diff
@@ -0,0 +1,133 @@
...
+///
+/// If there are no pending values when [_trigger] emits the first value on the
+/// source Stream will immediately flow through. Otherwise, the pending values
+/// and released when [_trigger] emits.
+///
+/// Errors from the source stream or the trigger are immediately forwarded to
+/// the output.
+class _Buffer<T, R> implements StreamTransformer<T, R> {
+  final Stream _trigger;
+  final _Collect _collect;
```
natebosch: I'm also trying to decide if we want to expose a version of this which takes `collect` as an argument. It doesn't match what I think people would expect from `buffer` and I can't think of a better name... `aggregate`?

Originally in `build_runner` I had built it taking an argument so I could aggregate into a map instead of a list, but changed it to be a closer match to Rx version of buffer.

Do you think it would be worthwhile to expose it?

jakemac53: I don't have a strong opinion, although if you did expose it then it might make sense to make it match the `fold` api, it is a similar concept.

Otherwise I think people will be confused about having to deal with a null value for `soFar`.

natebosch: I can't match `fold` precisely since I need a new `initialValue` after every trigger rather than 1 for the entire run. I could do `aggregate(Stream trigger, R initialValueFactory(), R collect(T value, R soFar))` but that seems noisy to me compared to handling null values inside `collect`. Would better documentation be enough?

jakemac53: Ah ya true :/. Good documentation would definitely help, or possibly just make the 2nd argument explicitly optional to indicate it may be null?

test/stream_transformers/buffer_test.dart
natebosch: done

lib/src/stream_transformers/buffer.dart
```diff
@@ -0,0 +1,133 @@
...
+  Stream<R> bind(Stream<T> values) {
+    StreamController<R> controller;
+    if (values.isBroadcast) {
+      controller = new StreamController<R>.broadcast();
+    } else {
+      controller = new StreamController<R>();
+    }
+
+    R currentResults;
+    bool waitingForValue = false;
```
natebosch: Good call, that's more clear.

lib/src/stream_transformers/buffer.dart
natebosch: Probably, I put the broadcast stream support in at the last minute and I think it's half baked.

Proper broadacast support isn't easy - I'm trying to decide if I want to work out all the edge cases now or ship without broadcast support at first and add it in later...

matanlurey:
Sorry a little late to the party - why isn't everything in `async/async.dart`?

The other top-level libraries are labeled deprecated for this exact reason.

natebosch:
@matanlurey - we were talking in the office yesterday about how to organize this. I had a few concerns about cramming these in with the overall async library.

- I expect we'll get at least 10 or so of these and I'm worried about how big async will get
- Most of async today is utilities at a different level of abstraction/usage than these. We'd see `buffer` used inside business logic, but `Delegating*` etc are a bit lower level. `StreamZip` is the same level though so async already has a range that captures what we're solving.

We were trying to decide between: {new package, async with a new library, async library} and a new library felt like the middle ground to me.

`stream_transformers` also doesn't fall into quite the same trap as the existing deprecated libraries, since those are very specific. I'm not creating a `buffer` library.

That said - I don't feel super strongly about it so if you think it is better to keep it all in a single top level library I'm open to arguments.

lib/src/stream_transformers/buffer.dart
jakemac53: I don't think its super important to support, users can always call `asBroadcastStream` on the resulting stream right?

natebosch: I think they would be able to. I'm not sure if there are any pitfalls in that scenario.

jakemac53:
There are advantages to having more than one public dart library:

- It helps when exploring the api in dartdoc, as it breaks things out by public library.
- It will help with ddc because we will create separate modules per public dart library.

Personally I kind of like the pattern of exporting everything from `async.dart`, but then also providing more fine grained libraries if people want to use them. I know not everybody agrees though :).

matanlurey:
For me it's a consistency thing - we deprecated the other libraries specifically because of this and most of our other first-party libs (collection, etc) follow this pattern of a single import. If we are rolling that back, fine, but let's make sure we are consistent then across our first-party libs.

/cc @kevmoo

jakemac53:
Agreed that consistency is good, and in the past the answer has generally been to have a single import. That does have its own advantages (you always know what to import for instance).

However, I think it is worth re-evaluating with modular compilers (like dartdevc) in mind.

matanlurey:
Sure. Just let's re-evaluate for all of dart-lang, not this specific lib :)

lib/src/stream_transformers/buffer.dart
natebosch: Using hasListeners for that.

test/stream_transformers/buffer_test.dart
natebosch: Looping around all combinations of broadcast/single subscription and running all of the tests now.

natebosch:
After some discussion we're going to start with this in a new package and defer the decision of putting them in package:async

PR #30, opened 2017-04-29
Travis

kevmoo:


lrhn:
Update the minor version number in pubspec.yaml too, otherwise LGTM.

PR #38, opened 2017-09-21
Skip typed_wrapper/stream_test

### LINKED ISSUE #35: TypeSafeStream<int> from Stream<Object> failing in Dart 1.25.0-dev.10.0
With Dart version 1.25.0-dev.10.0, some conversions from
Stream<Object> to TypeSafeStream<int> are failing.

If this is not the intent of the test, and the conversion should
succeed, it needs to be fixed in the library.

Filed a CL that fixes the test by changing some Stream<Object> to Stream<int>.

nex3:
We can't test this in 2.0 runtime semantics until test supports a
2.0-compatible platform (dart-lang/test#414).

Closes #35

### DIFF
--- a/test/typed_wrapper/stream_test.dart
+++ b/test/typed_wrapper/stream_test.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Skip("Re-enable this when test can run DDC (test#414).")
+
 import 'dart:async';
 
 import "package:async/src/typed/stream.dart";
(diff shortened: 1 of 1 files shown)


test/typed_wrapper/stream_test.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Skip("Re-enable this when test can run DDC (test#414).")
+
```
lrhn: Please add a library tag, so the annotation applies to that, not the import below. I know that it "works", but it's still ugly. 

(We should allow an empty `library;` declaration for that use-case).

nex3: That's contrary to the style of every other package that uses test annotations.

lrhn: Keep it then, until we make `library;` possible.

PR #43, opened 2018-01-02
Update for Dart 2.0 changes to Timer

keertip:


pubspec.yaml
```diff
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 description: Utility functions and classes related to the 'dart:async' library.
 homepage: https://www.github.com/dart-lang/async
 environment:
-  sdk: ">=1.22.0 <2.0.0"
+  sdk: ">=2.0.0-dev <2.0.0"
```
keertip: @kevmoo, does this look ok or should it be dev15.0 and above?

kevmoo: I think this should be based on the last dev release, yes!

leafpetersen: I think the syntax is something like sdk: '>=2.0.0-dev.15.0 <2.0.0 .

I think per our agreed process, adding this constraint is optional since the class is documented as "implementable".

keertip: This takes care of the problem that this version of the async package works only with pub from dev 15.0 due to constraints of it's dependencies. So it's not just for the Dart 2.0 changes.

kevmoo:
Yup yup

PR #44, opened 2018-01-05
Fix a StreamQueue bug

nex3:
Previously, StreamQueueTransaction assumed that its request was the
oldest request in the queue at the time at which it was committed or
rejected. This assumption wasn't always correct, and this change
avoids making it.


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.13.2
+
+* Fix a bug in `StreamQueue.startTransaction()` and related methods when
+  rejecting a transaction that isn't the oldest request in the queue.
+
```
lrhn: Ah yes, makes sense - you can reject a transaction before it gets its first event.

lib/src/stream_queue.dart
```diff
@@ -401,7 +401,7 @@ abstract class StreamQueue<T> {
 
   // ------------------------------------------------------------------
   // Methods that may be called from the request implementations to
-  // control the even stream.
+  // control the event stream.
```
lrhn: Thanks :)

PR #45, opened 2018-01-31
Updates for Dart 2.0 core library changes (wave 2.2)

leafpetersen:
This makes the initial changes necessary to make this package compatible with the corelib wave 2.2. changes.  See https://github.com/dart-lang/sdk/issues/31847 for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,10 @@
 * Add support for Dart 2.0 library changes to Stream and StreamTransformer.
   Changed classes that implement `StreamTransformer` to extend
   `StreamTransformerBase`, and changed signatures of `firstWhere`, `lastWhere`,
-  and `singleWhere` on classes extending `Stream`.  See also
-  https://github.com/dart-lang/sdk/issues/31847
+  and `singleWhere` on classes extending `Stream`.  See
+  also [issue 31847][sdk#31847].
+
+  [sdk#31847]: https://github.com/dart-lang/sdk/issues/31847
 
 ## 2.0.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.0.4
+
+* Add support for Dart 2.0 library changes to Stream and StreamTransformer.
+  Changed classes that implement `StreamTransformer` to extend
+  `StreamTransformerBase`, and changed signatures of `firstWhere`, `lastWhere`,
+  and `singleWhere` on classes extending `Stream`.  See also
+  https://github.com/dart-lang/sdk/issues/31847
+
 ## 2.0.3
 
 * Fix a bug in `StreamQueue.startTransaction()` and related methods when
--- a/lib/src/typed/stream.dart
+++ b/lib/src/typed/stream.dart
@@ -56,15 +56,15 @@ class TypeSafeStream<T> extends Stream<T> {
   Future<T> firstWhere(bool test(T element),
           {Object defaultValue(), T orElse()}) =>
       _stream.firstWhere(_validateType(test),
-          defaultValue: defaultValue ?? orElse);
+          defaultValue: defaultValue, orElse: orElse);
 
   Future<T> lastWhere(bool test(T element),
           {Object defaultValue(), T orElse()}) =>
       _stream.lastWhere(_validateType(test),
-          defaultValue: defaultValue ?? orElse);
+          defaultValue: defaultValue, orElse: orElse);
 
   Future<T> singleWhere(bool test(T element), {T orElse()}) async =>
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 2.0.4
+
+* Add support for Dart 2.0 library changes to Stream and StreamTransformer.
+  Changed classes that implement `StreamTransformer` to extend
+  `StreamTransformerBase`, and changed signatures of `firstWhere`, `lastWhere`,
+  and `singleWhere` on classes extending `Stream`.  See also
+  https://github.com/dart-lang/sdk/issues/31847
```
nex3: Nit: I'd write this

```md
* ... See also [issue 31847][sdk#31847].

[sdk#31847]: https://github.com/dart-lang/sdk/issues/31847
```

lib/src/typed/stream.dart
```diff
@@ -53,14 +53,18 @@ class TypeSafeStream<T> extends Stream<T> {
...
-
-  Future lastWhere(bool test(T element), {Object defaultValue()}) =>
-      _stream.lastWhere(_validateType(test), defaultValue: defaultValue);
-
-  Future<T> singleWhere(bool test(T element)) async =>
-      (await _stream.singleWhere(_validateType(test))) as T;
+  Future<T> firstWhere(bool test(T element),
+          {Object defaultValue(), T orElse()}) =>
+      _stream.firstWhere(_validateType(test),
+          defaultValue: defaultValue ?? orElse);
```
nex3: I think this should be `defaultValue: defaultValue, orElse: orElse`. I don't want this class to have *any* logic beyond what's necessary to get the types to work out.

leafpetersen: I want to (need to?) land this before @lrhn lands the breaking change in the SDK, so currently there is no `orElse` on stream to pass this to.  I will file an issue to follow up with a PR removing the defaultsTo argument and this logic here after the SDK changes land.

nex3: My understanding from talking with @srawlins at Dart Conf was that we were just going to have this package briefly be in a broken state on master.

CHANGELOG.md
leafpetersen: Done.

lib/src/typed/stream.dart
leafpetersen: Ah, I see.  Yes, I'm fine with that solution if you prefer.  It does reduce the number of PRs.

leafpetersen: @keertip @srawlins If I make all of the changes in one go here and leave it on a branch, does that work for testing internal code?  I think that means you would have to test with an atomic CL, since you couldn't roll this package before you roll the SDK.  You would then presumably also have to roll this package with the SDK when we finally land.  Does that work, or should we do these changes in two passes?

srawlins: Yeah that works with testing internal code. It... should work for SDK as well, right? `orElse` can land, and aync's version can bump in DEPS, all in the same changelist.

leafpetersen: Thinking about it, I guess I could push multiple commits to this branch, and then we could roll internal to the first tagged commit (this version), then roll internal to the second tagged commit (the final version).  As far as this package is concerned, there is just one PR with a bunch of commits, but internally we can point to different hashes in this branch at different points in the process.

nex3: Do we need multiple commits? Can't both the SDK and internal atomically add `orElse` and bump their dependency on `async` in the same commit?

leafpetersen: We want internal to be green before we land the changes in the SDK.

To get internal green, we need to run presubmits, fix, iterate.

It makes running presubmits easier if we can roll a dev.20 compatible version internally.

So I think having two commits to reference here for now would be better.  Since we're referencing them off the branch, I don't think it should matter in the long run - we'll just leave this branch sitting here for now, and land the PR atomically as usual once everything is ready.

nex3: :+1:

jakemac53:
2.0.0-dev.23.0 is now released on windows so any ddc builds depending on this package are now broken at the latest dev.

I will try pushing a commit with desired changes and we can get it merged/published?

jakemac53:
@nex3 @srawlins  can you take a look? I believe I made the desired changes.

I don't actually see the linux build pushed yet though so I couldn't validate yet.

leafpetersen:
Your changes LGTM.

nex3:
I've verified locally that tests and analysis pass with dart-lang/sdk@365f7b5a8b6ef900a5ee23913b7203569b81b175.

Issue #48, opened 2018-02-24
Run tests with DDC


natebosch:
There are some potential Strong mode issues within this package. We should run tests with DDC.

matanlurey:
This is long done.

matanlurey:
Er, it is not, on travis. Oops!

natebosch:
The main goal here was to validate Dart 2 runtime. Now that the VM has a Dart 2 runtime I think we can drop this.

PR #50, opened 2018-02-24
Add deps for build_runner

natebosch:
Towards #48

Allows us to run `pub run build_runner test -- -p chrome` to test with
DDC. Exposes issues with the package and with tests so not enabling on
Travis for now.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,6 +11,7 @@ dev_dependencies:
   fake_async: "^0.1.2"
   stack_trace: "^1.0.0"
   test: "^0.12.0"
+  # For building and testing with DDC
   build_runner: ^0.7.11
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -11,3 +11,6 @@ dev_dependencies:
   fake_async: "^0.1.2"
   stack_trace: "^1.0.0"
   test: "^0.12.0"
+  build_runner: ^0.7.11
```
kevmoo: I'd add a `# note` here about why these are included – without other documentation or usage, it's not clear

nex3:
I'm :-1: on this change. I'm close to having `test` support DDC natively, and I'd prefer to solve #48 that way rather than using the build system.

Issue #49, opened 2018-02-24
Drop dependency on fake_async


natebosch:
Pull requests on that package are getting ignored. That package does not compile with DDC.

We either need to fork that package or drop our dependency.

cc @kevmoo 

This  is a blocker for #48

nex3:
`fake_async` is important for race-condition-free testing of asynchrony, and I'm happy to maintain it. I'll start the process by reaching out to Sean Eagan.

PR #52, opened 2018-02-28
Run tests in DDC w/ incremental caching

matanlurey:
Taken from here:
https://github.com/dart-lang/mockito/blob/master/.travis.yml

Example builds/tests are super fast:
https://travis-ci.org/dart-lang/mockito/builds/347054827


matanlurey:
Well good news - this catches that `fake_async` is broken.

What are we doing here? @srawlins @natebosch @munificent

natebosch:
https://github.com/dart-lang/async/issues/49

I think we're going to need to make a decision whether we want to take over the fake_async package since the author is not active.

cc @nex3 @kevmoo

matanlurey:
Can we temporarily use Quiver's FakeAsync (the original)? I think it works.

matanlurey:
Ping @nex3 @natebosch @kevmoo

natebosch:
I think we have permission to take over package:fake_async and package:clock.

Our plan is to move the code from quiver to these packages. We want to have only 1 copy and the ones in quiver have more users so we're better off starting from that. Use an `export` to make the migration to that import easier/incremental.

I don't know if we have anyone slotted to take on the work though. In practice that does mean migrating to the package:quiver version of this class - but if we can do it without the (temporary) import to quiver that would be optimal.

nex3:
I'm happy to do the migration.

natebosch:
I think that the `fake_async` and `clock` stuff has all been migrated? At least the tests seem to be passing in Dart 2 runtime.

I don't think that there is still value in this change now that we're getting Dart 2 runtime testing in the VM. @matanlurey - please reopen if you disagree and still want to pursue DDC tests in this repo.

PR #55, opened 2018-03-13
Delete typed future, use Dart 2 instead.

matanlurey:
I'd love to do this with `StreamSubscription` too, but no option there :-/


lrhn:
Update the version number to 2.0.7 (or 2.0.7-dev if you plan to land more changes before releasing).
Otherwise LGTM.

PR #56, opened 2018-04-11
Remove upper case constants

lrhn:


lrhn:
What should the changelog say? Just "Updated SDK version to ...", or should it say why?

lrhn:
PTAL

PR #63, opened 2018-06-22
Enable prefer_typing_uninitialized_variables

natebosch:
This helps rule out some possibilities for the new Dart 2 failure and is
good practice.

Also:
- Update dependencies on build and test packages.
- Ignore todo from the analysis server.
- Move off of deprecated APIs from test.


natebosch:
Note that Travis will still be failing, I have not found the failure cause yet.

PR #65, opened 2018-06-29
Fix places awaiting void, for dart 2.

MichaelRFairhurst:
Looks like cache.invalidate does do some asynchronous stuff
(`splitter.close()`, but notably not `_stale.close()` which returns void).
Return the future for that stuff without changing its order of
operations.

Also some place awaiting `expect()` which appeared to be intended to be
awaiting `cancel()`, switching those still passes tests.

### DIFF
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -89,12 +89,13 @@ class AsyncCache<T> {
   }
 
   /// Removes any cached value.
-  void invalidate() {
+  Future invalidate() {
     _cachedValueFuture = null;
-    _cachedStreamSplitter?.close();
+    Future invalidate = _cachedStreamSplitter?.close();
     _cachedStreamSplitter = null;
     _stale?.cancel();
     _stale = null;
+    return invalidate;
   }
 
   void _startStaleTimer() {
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -37,19 +37,19 @@ class AsyncCache<T> {
   /// Fires when the cache should be considered stale.
   Timer _stale;
 
-  /// Creates a cache that invalidates after an in-flight request is complete.
-  ///
-  /// An ephemeral cache guarantees that a callback function will only be
-  /// executed at most once concurrently. This is useful for requests for which
-  /// data is updated frequently but stale data is acceptable.
-  factory AsyncCache.ephemeral() => new AsyncCache(Duration.zero);
-
   /// Creates a cache that invalidates its contents after [duration] has passed.
   ///
   /// The [duration] starts counting after the Future returned by [fetch]
   /// completes, or after the Stream returned by [fetchStream] emits a done
   /// event.
   AsyncCache(this._duration);
+
+  /// Creates a cache that invalidates after an in-flight request is complete.
+  ///
+  /// An ephemeral cache guarantees that a callback function will only be
+  /// executed at most once concurrently. This is useful for requests for which
+  /// data is updated frequently but stale data is acceptable.
+  factory AsyncCache.ephemeral() => new AsyncCache(Duration.zero);
 
   /// Returns a cached value from a previous call to [fetch], or runs [callback]
   /// to compute a new one.
@@ -89,13 +89,16 @@ class AsyncCache<T> {
   }
 
   /// Removes any cached value.
-  Future invalidate() {
+  Null invalidate() {
     _cachedValueFuture = null;
-    Future invalidate = _cachedStreamSplitter?.close();
+    // TODO: This does not await, but probably should.
+    _cachedStreamSplitter?.close();
     _cachedStreamSplitter = null;
     _stale?.cancel();
     _stale = null;
-    return invalidate;
+
+    // TODO: This does not return a future, but probably should.
+    return null;
   }
 
   void _startStaleTimer() {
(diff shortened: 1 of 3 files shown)


nex3:
I'm not really actively maintaining this package anymore.

lib/src/async_cache.dart
```diff
@@ -89,12 +89,13 @@ class AsyncCache<T> {
   }
 
   /// Removes any cached value.
-  void invalidate() {
+  Future invalidate() {
     _cachedValueFuture = null;
-    _cachedStreamSplitter?.close();
+    Future invalidate = _cachedStreamSplitter?.close();
```
kevmoo: I'd just make the function async and await here – in theory you could be avoiding some unneeded async silly if the splitter != null – but you are also adding some complexity to how this method reads.

Maybe add an await and put the null check in a block – best of both worlds?

kevmoo: Also this is a ~breaking change – folks will start getting warnings about awaiting here if they have the lint enabled.

Why is this change needed?

MichaelRFairhurst: Well, technically I would say this is a bugfix.

If the invalidate method does asynchronous work, and the future of that work is thrown away, its impossible to await invalidation. There is at least some code (in the tests) awaiting validation, and that silently did not actually await the invalidation.

However, this could be maintained as returning void, and we could add a `invalidateAsync` method which returns the future of `_cachedStreamSplitter?.close()` ?

Regarding the order of operations, I just really wanted to land a change that's as small as possible -- so this won't add asychrony in the void case (like before) and it won't change the order of operations vs before. Not necessarily important, and not as clean as `await`ing here, but I just didn't want spooky asynchrony changes to stall the CL.

If you aren't spooked by the await I'll do that though! And do you want it to be a new method? The good news is that dart 2 will not let them use the wrong method ("you can't await invalidate() as it returns void")

MichaelRFairhurst: ping @kevmoo

lib/src/async_cache.dart
```diff
@@ -89,12 +89,13 @@ class AsyncCache<T> {
   }
 
   /// Removes any cached value.
-  void invalidate() {
+  Future invalidate() {
```
natebosch: how about `Future<void>`?

lib/src/async_cache.dart
natebosch: I don't understand how this relates to an `await void`? Was something in the tests awaiting the result of this function before? Wasn't that effectively `await null`? Can it still be?

MichaelRFairhurst: Yeah, a test was awaiting this method. It *could* definitely return null, but I would consider the following a bug:

```
Future f() {
  spawn_async_operation_without_await();
  return null;
}
```

"function returning future doesn't actually block until work is complete" is the title I'd give that bug!

Alternatively, and I mean this as a strawman but I also honestly prefer it, I could do

```
Null f() {
  spawn_async_operation_without_await();
  return null;
}
```

lib/src/async_cache.dart
MichaelRFairhurst: Thanks, definitely should do that..!

lib/src/async_cache.dart
natebosch: What I mean is, the old behavior was already translatable to `await null;` This isn't a case of a `Future` slipping through a `void` signature and being an otherwise valid `await`. This is a case of a `Null` going through a `void` signature. I agree on principle with the idea of making this async work _possible_ to await. But as it is this change is not maintaining the same semantics in the test.

MichaelRFairhurst: Yeah, I was careful to preserve the ordering semantics just out of conservatism, but didn't do so for this part, which is arguably inconsistent. I can leave a TODO and return null.

lib/src/async_cache.dart
```diff
@@ -89,12 +89,16 @@ class AsyncCache<T> {
...
-  void invalidate() {
+  Null invalidate() {
     _cachedValueFuture = null;
+    // TODO: This does not await, but probably should.
     _cachedStreamSplitter?.close();
     _cachedStreamSplitter = null;
     _stale?.cancel();
     _stale = null;
+
+    // TODO: This does not return a future, but probably should.
```
natebosch: I think we can leave this method as it was (although we can either file an issue or add a TODO).

This can continue to be a `void` return and we can rewrite `await foo.invalidate();` to

```dart
foo.invalidate();
await null;
```

natebosch: Ah, missed your commend that you had explicitly avoided this. FWIW I prefer it since it localizes the "fix" to where the await is. It's also slightly less "breaking" to later change from `void` to `Future<void>` than it would be to change from `Null` to `Future<void>`.

MichaelRFairhurst: Part of the problem is that if this code *should* be awaited, because it *does* do asynchronous work, then I don't want to migrate all code awaiting this to no longer do so, only to then fix it and have to migrate all other code back..

natebosch: We've decided to keep allowing `await void` for now right? Maybe we should punt on this and file an issue to do the proper fix of making it return a `Future`?

MichaelRFairhurst: We might still get this in in time. But its not a priority

PR #66, opened 2018-07-13
Add missing await to cache invalidation

MichaelRFairhurst:


lib/src/async_cache.dart
```diff
@@ -89,9 +89,9 @@ class AsyncCache<T> {
   }
 
   /// Removes any cached value.
-  void invalidate() {
+  Future<void> invalidate() async {
     _cachedValueFuture = null;
-    _cachedStreamSplitter?.close();
+    await _cachedStreamSplitter?.close();
     _cachedStreamSplitter = null;
     _stale?.cancel();
```
natebosch: can we move the `_stale?.cancel()` above the `await` since it might otherwise fire while we're waiting for the `close()`?

kevmoo: also, you want to do the null check separately. Otherwise you'll waste an async op awaiting null

mit-mit:
Closing as assumed stale

PR #68, opened 2018-09-14
Fix CancelableOperation.valueOrCancellation's type signature

ds84182:
Otherwise it's annoying to deal with in Dart 2.0.


lrhn:
LGTM if you update the version number and changelog.

This is a close to a non-breaking change that a Dart 2 code change can get, but I guess we should increment the minor version number of the package so people can request this change.

ds84182:
Sorry about the delay, busy week!

Just updated the pubspec and the changelog

PR #71, opened 2018-12-11
dev dep: latest build_web_compilers

kevmoo:


natebosch:
I'm not sure if there is still value in testing with DDC this was last updated when VM didn't support strong mode. Any thoughts on testing with VM only?

kevmoo:
> I'm not sure if there is still value in testing with DDC this was last updated when VM didn't support strong mode. Any thoughts on testing with VM only?

🤷‍♂️ @matanlurey did the work here. Thoughts?

kevmoo:
Going to land this just to get to latest. We can chat about removing the whole thing here and in pkg:collection on a separate thread...

PR #73, opened 2019-01-25
Merge _StreamQueue into StreamQueue

natebosch:
The original reason to split into an abstract and concrete class was to
allow other implementations that would allow for the transaction
implementation. Since that was implemented without a separate subclass
merge the implementation back to one class to avoid unnecessary
abstraction and make it easier to debug.


natebosch:
Was originally split in https://codereview.chromium.org//1305063002

lrhn:
LGTM.

PR #81, opened 2019-03-28
Add CancelableOperation.then

### LINKED ISSUE #80: Implement CancelableOperation.then
It would be useful to be able to transform `CancelableOperation` in the same way as a `Future`. The API I envisage is:

```
CancelableOperation<R> then<R>(CancelableOperation<R> Function(T) onValue,
    {Function onError});

CancelableOperation<R> thenFuture<R>(FutureOr<R> Function(T) onValue,
    {Function onError});
```

This should behave similarly to `Future.then`, but with the following extras:

1. If the original operation is canceled then the returned operation is canceled.
2. If the operation returned by `onValue` is canceled then the returned operation is canceled.
3. If the returned operation is canceled then the original operation is canceled or the `onValue` operation is canceled if it has been started.

TastyPi:
Fixes #80.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -75,6 +75,49 @@ class CancelableOperation<T> {
     return completer.future;
   }
 
+  /// Registers callbacks to be called when this operation completes.
+  ///
+  /// Behaves similarly to [Future.then] but with the following cancellation
+  /// behaviour:
+  /// - If this operation is canceled then the result operation is canceled.
+  /// - If the operation created by [onValue] or [onError] is canceled then the
+  ///   result operation is canceled.
+  /// - If the result operation is canceled then the operation created by
+  ///   [onValue] or [onError] is canceled if one has been called otherwise this
+  ///   operation is canceled.
+  CancelableOperation<R> then<R>(CancelableOperation<R> Function(T) onValue,
+      {CancelableOperation<R> Function(Object, StackTrace) onError}) {
+    CancelableOperation<R> nextOperation;
+
+    // Cancel this operation or the nextOperation if the result operation is
+    // canceled.
+    Future onCancel() =>
+        nextOperation == null ? cancel() : nextOperation.cancel();
+
+    final completer = CancelableCompleter<R>(onCancel: onCancel);
+
+    FutureOr cancelThen() {
+      // Prevent loops.
+      if (!completer.isCanceled) {
+        return completer._cancel();
+      }
+    }
+
+    Future<R> runNextOperation(CancelableOperation<R> operation) {
+      nextOperation = operation;
+      nextOperation._canceled.whenComplete(cancelThen);
+      return nextOperation.value;
+    }
+
+    completer.complete(value.then((v) => runNextOperation(onValue(v)),
+        onError: onError != null
+            ? (e, s) => runNextOperation(onError(e, s))
+            : null));
+
+    _canceled.whenComplete(cancelThen);
+    return completer.operation;
+  }
+
   /// Cancels this operation.
   ///
   /// This returns the [Future] returned by the [CancelableCompleter]'s
@@ -86,6 +129,8 @@ class CancelableOperation<T> {
 
   /// Whether this operation completed before being canceled.
   bool get isCompleted => _completer.isCompleted;
+
+  Future<void> get _canceled => _completer._cancelMemo.future;
 }
 
 /// A completer for a [CancelableOperation].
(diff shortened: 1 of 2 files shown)


natebosch:
@lrhn - any thoughts here?

lrhn:
I can see why this could be convenient. Not sure it's worth the complexity, but at least the complexity is localized to the new method.

This essentially creates a single cancelable operation which executes one operation first, and when that completes, then executes the other. The combined operation acts like whichever operation is currently running wrt. canceling. 

I'm a little worried that someone *else* canceling one of the operations will also cancel the combined operations. That suggests that you share the cancelable operations among different pieces of code, and either one can cancel the operation. That seems unsafe. I'd probably *assume* ownership of the operation, and therefore assume that nobody else cancels the operation—or if they do, treat that as just the way that operation completed. I don't think propagating cancels upwards is necessary, or necessarily desirable (and that's also what requires access to a private member of the foreign operation).
Maybe instead have an `CancelableOperation<R> onCancel()` argument that you call, then you can decide exactly what to do if the first operation was cancelled (and the first operation is yourself, so you have access to the private implementation of that).

The change also needs to update `CHANGLOG.md` and `README.md`, and update the version number in `pubspec.yaml`.

Be aware that changing any API is a breaking change, so this should require a major version increment.
(You can argue that *most likely* nobody currently implements `CancelableOperation`, in which case it can be defended as a minor version increment).

TastyPi:
I'll start with a disclaimer: I don't actually have a use for the `then` method, I only actually want the `thenFuture` method so I can easily transform a `CancelableOperation<A>` into a `CancelableOperation<B>`. So if `then` is controversial I'd be happy with just adding `thenFuture`, it just seemed weird to me from an API perspective to have `thenFuture` without `then`.

Back to `then`, here's my thinking for the cancelling behaviour:

1. If the original operation is canceled then the intermediate operation has not been created yet so can be ignored, and the resulting operation should be canceled as it will never end.
2. If the intermediate operation is canceled then the original operation by definition has finished so can be ignored, and the resulting operation should be canceled as it will never end. Note that it can only be canceled if the callback that created it keeps a reference to it and cancels it, which would be weird to do but probably something that should be supported.

I think these are uncontroversial, but let me know if you don't think so. It should be possible to reimplement `then` using `valueOrCancellation` if you're concerned about using private members.

3. If the resulting operation is canceled what should happen?

It sounds like this is where the debate is. My thinking for this method is that most people will want to use it like this:

```dart
final operation = doCancelableThing().then(doOtherCancelableThing);
```

In this scenario I think it is clear the user would want the original operation to be canceled, that being said I can also imagine cases where you don't want the original to be canceled (e.g. doing multiple things on completion). With your proposal of adding `onCancel` this could be done like so:

```dart
final firstOperation = doCancelableThing();
final operation = firstOperation
    .then(doOtherCancelableThing, onCancel: firstOperation.cancel);
```

(I don't think `onCancel` should return `CancelableOperation` like you suggested as if the resulting operation is canceled it would be weird for it to not be canceled but be replaced with a new operation. I think using `FutureOr Function() onCancel` would make more sense and that matches other `onCancel` parameters.)

This works, but I don't like that chaining doesn't work for what I think will be the more common use case. Maybe add a `bool cancelThisOnCancel = false` named parameter so you can do:

```dart
final operation = doCancelableThing()
    .then(doOtherCancelableThing, cancelThisOnCancel = true);
```

I'm open to other naming suggestions for the parameter, but I think this makes it clear at the call site whether the original is being canceled and supports nice chaining.

Another alternative would be for `onCancel` to take the currently executing `CancelableOperation` as an argument (i.e. the original or one of the intermediate ones) and do this:

```dart
final operation = doCancelableThing()
    .then(doOtherCancelableThing, onCancel: (o) => o.cancel());
```

Or maybe `FutureOr<bool> Function() onCancel` where returning `true` propagates the cancel:

```dart
final operation = doCancelableThing()
    .then(doOtherCancelableThing, onCancel: () => true);
```

natebosch:
> I don't actually have a use for the `then` method, I only actually want the `thenFuture` method so I can easily transform a `CancelableOperation<A>` into a `CancelableOperation<B>`. So if `then` is controversial I'd be happy with just adding `thenFuture`, it just seemed weird to me from an API perspective to have `thenFuture` without `then`.

FWIW I'd be in favor of keeping the behavior from `theFuture` and naming it `then`.

If we don't have a known use case today for following up with another `CancelableOperation` and treating the cancellation this way then I'd prefer to avoid it altogether and save the nicer name for the use case which seems more likely. We can always add a `.thenCancelable` later if we need it and save all the discussion around behavior for if and when that need arises.

TastyPi:
Sounds good to me, I'll create a new pull request with just then.

PR #83, opened 2019-04-04
Implement CancelableOperation.then

### LINKED ISSUE #80: Implement CancelableOperation.then

TastyPi:
Fix #80.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.2.0
+
+* Add `then` to `CancelableOperation`.
+
 ## 2.1.0
 
 * Fix `CancelableOperation.valueOrCancellation`'s type signature
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -75,6 +75,47 @@ class CancelableOperation<T> {
     return completer.future;
   }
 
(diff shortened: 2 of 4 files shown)


TastyPi:
@natebosch @lrhn 

The issue of whether to propagate the cancellation to the original operation still happens here, but I think this is a lot clearer with what is happening than the version with callbacks returning CancelableOperations.

lrhn:
LGTM. It's a much more focused API than the original one. A cancelable operation is a combination of result future and cancel channel. It makes sense to have an operation which acts on both (a `then` on the future, plus back-propagation of cancel).

PR #86, opened 2019-05-23
run dartfmt to fix travis

srawlins:
I'm confused. I think maybe I submitted directly to master??? And broke travis because of fmt?

So this... fixes that?


natebosch:
Can you make sure the commit message reflects the change - run dartfmt to fix travis, and not share the same title as the previous commit?

PR #89, opened 2019-07-16
Implement RestartableTimer.tick

### LINKED ISSUE #88: RestartableTimer#tick not implemented
Hello, today I read the official documentation and found that [Timer](https://api.flutter.dev/flutter/dart-async/Timer-class.html) has an implementation class [RestartableTimer](https://api.flutter.dev/flutter/package-async_async/RestartableTimer-class.html), and then try to use it, I imported dart:async, because Timer is inside the dart:async package. But can't find the RestartableTimer class. Why not add the RestartableTimer to the dart:async package?
Also, the get [tick](https://api.flutter.dev/flutter/package-async_async/RestartableTimer/tick.html) of RestartableTimer will throw an exception, I found it based on the comment. But Timer's get tick is normal.

[RestartableTimer#tick](https://api.flutter.dev/flutter/package-async_async/RestartableTimer/tick.html)
```
@override
// TODO: Dart 2.0 requires this method to be implemented.
// See https://github.com/dart-lang/sdk/issues/31664
// ignore: override_on_non_overriding_getter
int get tick {
  throw UnimplementedError("tick");
}
```

natebosch:
Closes #88

Extra cleanup:
- change a constructor body to an initializer list
- remove an `@override` annotation

### DIFF
--- a/lib/src/restartable_timer.dart
+++ b/lib/src/restartable_timer.dart
@@ -44,11 +44,11 @@ class RestartableTimer implements Timer {
     _timer.cancel();
   }
 
+  /// The number of durations preceding the most recent timer event on the most
+  /// recent countdown.
+  ///
+  /// Calls to [reset] will also reset the tick so subsequent tick values may
+  /// not be strictly larger than previous values.
   @override
-  // TODO: Dart 2.0 requires this method to be implemented.
-  // See https://github.com/dart-lang/sdk/issues/31664
-  // ignore: override_on_non_overriding_getter
-  int get tick {
-    throw UnimplementedError("tick");
-  }
+  int get tick => _timer.tick;
 }
--- a/lib/src/restartable_timer.dart
+++ b/lib/src/restartable_timer.dart
@@ -25,9 +25,8 @@ class RestartableTimer implements Timer {
   ///
   /// The [callback] function is invoked after the given [duration]. Unlike a
   /// normal non-periodic [Timer], [callback] may be called more than once.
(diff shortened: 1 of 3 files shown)


natebosch:
Adding @lrhn in case there are any concerns since we're breaking the monotonic increase expectations. I don't think it's worth trying to do something more complex, though we _could_ consider an implementation that keeps that guaranteed by keeping track of the `tick` when we `reset()` and adding to the new timer's tick..

lrhn:
No concerns. A reset timer works like a *new* timer, and the tick of a non-repeating timer is always 0 (before firing) or 1 (after firing) anyway.

lib/src/restartable_timer.dart
```diff
@@ -44,11 +44,11 @@ class RestartableTimer implements Timer {
     _timer.cancel();
   }
 
+  /// The number of durations preceding the most recent timer event on the most
+  /// recent countdown.
+  ///
+  /// Calls to [reset] will also reset the tick so subsequent tick values may
+  /// not be strictly larger than previous values.
   @override
```
lrhn: Do remove the `@override` while you are here. We don't use `@override` in `package:async`. (This is the only instance of `@override` in the entire package).

natebosch: Sure, I can for now.

I think we should strongly consider adding them everywhere in this package and enabling the lint. We implement a lot of interfaces and it's really nice to be able to see quickly which are the new methods.

PR #90, opened 2019-10-03
Add StreamGroup.mergeBroadcast

natebosch:


test/stream_group_test.dart
```diff
@@ -435,6 +435,23 @@ main() {
...
+        StreamGroup.mergeBroadcast([controller1.stream, controller2.stream]);
+
+    controller1.add("first");
+    controller1.close();
+    controller2.add("second");
+    controller2.close();
+
+    expect(merged.isBroadcast, isTrue);
+
+    expect(merged.toList(), completion(unorderedEquals(["first", "second"])));
```
jakemac53: does this need to be an `unorderedEquals` check? In theory these would always come in the order they are emitted by the given streams right?

natebosch: 🤷‍♂ - it duplicates the existing test... I'll followup on this.

PR #104, opened 2020-03-13
Expand docs for isCompleted

### LINKED ISSUE #102: CancelableOperation.isCompleted is True Before Future Completes
`CancelableOperation.isCompleted` returns `true` immediately after constructing the operation even if the future it was constructed with hasn't resolved. This was unexpected to me, and I assume this is not the intended behavior. Below is a little unit test that demonstrates the problem. The expectation on the final line of the test fails, though I would expect it to succeed.

```dart
import 'dart:async';

import 'package:test/test.dart';

void main() {
  group('CancelableOperation', () {
    test('isComplete should be false before completed', () {
      var completer = Completer<void>();
      var operation = CancelableOperation.fromFuture(completer.future);
      expect(operation.isCompleted, isFalse);
    });
  });
}
```

natebosch:
Closes #102

This field is confusing because it documents whether the backing
completer has had `complete` called which doesn't necessarily correspond
to when the value is available if it was completed with a `Future`.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -30,9 +30,10 @@ class CancelableOperation<T> {
   /// [onCancel] will be called synchronously when the operation is canceled.
   /// It's guaranteed to only be called once.
   ///
-  /// This is the same as creating a [CancelableCompleter] and completing it
-  /// with [inner]. The [isCompleted] field will always be true, regardless of
-  /// whether the completer backing [inner] has completed.
+  /// Calling this constructor is equivalent to creating a [CancelableCompleter]
+  /// and completing it with [inner]. As such, [isCompleted] is true from the
+  /// moment this [CancelableOperation] is created, regardless of whether
+  /// [inner] has completed yet or not.
   factory CancelableOperation.fromFuture(Future<T> inner,
       {FutureOr Function() onCancel}) {
     var completer = CancelableCompleter<T>(onCancel: onCancel);
@@ -132,8 +133,9 @@ class CancelableOperation<T> {
   /// Whether the [CancelableCompleter] backing this operation has been
   /// completed.
   ///
-  /// This does not indicate that the completer backing [value] has been
-  /// completed if the [CancelableCompleter] was completed with a future.
+  /// This value being true does not imply that the [value] future has
+  /// completed, but merely that it is no longer possible to [cancel] the
+  /// operation.
   bool get isCompleted => _completer.isCompleted;
 }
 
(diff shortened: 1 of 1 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -29,6 +29,10 @@ class CancelableOperation<T> {
   ///
   /// [onCancel] will be called synchronously when the operation is canceled.
   /// It's guaranteed to only be called once.
+  ///
+  /// This is the same as creating a [CancelableCompleter] and completing it
+  /// with [inner]. The [isCompleted] field will always be true, regardless of
+  /// whether the completer backing [inner] has completed.
```
lrhn: Generally, avoid using an unqualified "this" (or "it" or similar nondescript words) in documentation. It's almost always better to qualify.

Maybe:
```dart
/// Calling this constructor is equivalent to creating a [CancelableCompleter] 
/// and completing it with [inner]. As such, [isCompleted] is true from the moment
/// this [CancelableOperation] is created, regardless of whether [inner] has completed 
/// yet or not.

We don't know whether there is a completer backing `inner`, so we can't refer to it. It's just a future, they can come from anywhere.

File: lib/src/cancelable_operation.dart [hunk 2]
```diff
@@ -125,7 +129,11 @@ class CancelableOperation<T> {
   /// Whether this operation has been canceled before it completed.
   bool get isCanceled => _completer.isCanceled;
 
-  /// Whether this operation completed before being canceled.
+  /// Whether the [CancelableCompleter] backing this operation has been
+  /// completed.
+  ///
+  /// This does not indicate that the completer backing [value] has been
```
lrhn:
```dart
/// This value being true does not imply that the [value] future has completed, 
/// but merely that it is no longer possible to [cancel] the operation.
```

Maybe this value is not that useful. I would personally prefer to not have `isCompleted` at all, or have named it `canCancel`. Probably not worth the breaking change, though.

lib/src/cancelable_operation.dart
natebosch: Done

lib/src/cancelable_operation.dart
natebosch: This raises a question - if it is not possible to cancel a `CancelableOperation.fromFuture`, why does that constructor exists? What do we get out of wrapping if we can't cancel?

lrhn: You get to pass a `Future` to something expecting a `CancelableOperation`. It's an adapter, partial because the adeptee doesn't support the full API it's being adapted for.

I would probably have allowed you to cancel until the future result was available, whether the underlying completer is completed or not, but that would require more complicated plumbing - not just completing the underlying operation/future completers with the future.

PR #108, opened 2020-03-17
Null safety migration

jakemac53:
~Tests not migrated yet, but sending this out for review in the meantime.~

Tests migrated as well now

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -35,7 +35,7 @@ class CancelableOperation<T> {
   /// moment this [CancelableOperation] is created, regardless of whether
   /// [inner] has completed yet or not.
   factory CancelableOperation.fromFuture(Future<T> inner,
-      {FutureOr Function() onCancel}) {
+      {FutureOr Function()? onCancel}) {
     var completer = CancelableCompleter<T>(onCancel: onCancel);
     completer.complete(inner);
     return completer.operation;
@@ -55,7 +55,7 @@ class CancelableOperation<T> {
     value.then((value) {
       controller.add(value);
       controller.close();
-    }, onError: (error, StackTrace stackTrace) {
+    }, onError: (Object error, StackTrace stackTrace) {
       controller.addError(error, stackTrace);
       controller.close();
     });
@@ -68,7 +68,10 @@ class CancelableOperation<T> {
   /// If this operation completes, this completes to the same result as [value].
   /// If this operation is cancelled, the returned future waits for the future
   /// returned by [cancel], then completes to [cancellationValue].
-  Future<T> valueOrCancellation([T cancellationValue]) {
+  ///
+  /// If [T] is not nullable then [cancellationValue] must provide a valid
+  /// value, otherwise it will throw on cancellation.
+  Future<T> valueOrCancellation([T? cancellationValue]) {
     var completer = Completer<T>.sync();
     value.then((result) => completer.complete(result),
         onError: completer.completeError);
@@ -93,8 +96,8 @@ class CancelableOperation<T> {
   /// If [propagateCancel] is `true` and the returned operation is canceled then
   /// this operation is canceled. The default is `false`.
   CancelableOperation<R> then<R>(FutureOr<R> Function(T) onValue,
-      {FutureOr<R> Function(Object, StackTrace) onError,
-      FutureOr<R> Function() onCancel,
+      {FutureOr<R> Function(Object, StackTrace)? onError,
+      FutureOr<R> Function()? onCancel,
       bool propagateCancel = false}) {
     final completer =
         CancelableCompleter<R>(onCancel: propagateCancel ? cancel : null);
@@ -109,7 +112,7 @@ class CancelableOperation<T> {
           completer._cancel();
         }
       }
-    }, onError: (error, StackTrace stackTrace) {
+    }, onError: (Object error, StackTrace stackTrace) {
       if (!completer.isCanceled) {
         if (onError != null) {
           completer.complete(Future.sync(() => onError(error, stackTrace)));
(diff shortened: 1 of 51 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -68,7 +68,10 @@ class CancelableOperation<T> {
   /// If this operation completes, this completes to the same result as [value].
   /// If this operation is cancelled, the returned future waits for the future
   /// returned by [cancel], then completes to [cancellationValue].
-  Future<T> valueOrCancellation([T cancellationValue]) {
+  ///
+  /// If [T] is not nullable then [cancellationValue] must provide a valid
```
natebosch: Should we be collecting any breaking changes we _should_ be making to improve ergonomics in an issue or something?

lib/src/result/error.dart
```diff
@@ -8,19 +8,19 @@ import 'result.dart';
...
 
 /// A result representing a thrown error.
-class ErrorResult implements Result<Null> {
+class ErrorResult implements Result<Never> {
   /// The error object that was thrown.
   final Object error;
 
   /// The stack trace corresponding to where [error] was thrown.
-  final StackTrace stackTrace;
+  final StackTrace? stackTrace;
```
natebosch: This should be non-null and we should handle the defaulting if we need to.

lib/src/result/error.dart
```diff
@@ -8,19 +8,19 @@ import 'result.dart';
...
 
   @override
   bool get isValue => false;
   @override
   bool get isError => true;
   @override
-  ValueResult<Null> get asValue => null;
+  ValueResult<Never>? get asValue => null;
```
natebosch: Another to collect as a breaking change we should make.

lib/src/result/result.dart
```diff
@@ -64,7 +64,7 @@ abstract class Result<T> {
     try {
       return ValueResult<T>(computation());
     } catch (e, s) {
-      return ErrorResult(e, s);
+      return ErrorResult(e as Object, s);
```
natebosch: https://github.com/dart-lang/sdk/issues/41074

Looks like `on Object catch` might be the preference.

lib/src/result/result.dart
```diff
@@ -108,15 +108,15 @@ abstract class Result<T> {
...
+            completer.complete(results.cast<Result<T>>().toList());
           }
         });
       } else {
         results.add(Result<T>.value(element as T));
       }
     }
     if (pending == 0) {
-      return Future<List<Result<T>>>.value(results);
+      return Future.value(results.cast<Result<T>>().toList());
```
natebosch: I've been doing `List.from(results)` instead.

lib/src/stream_completer.dart
```diff
@@ -173,8 +173,9 @@ class _CompleterStream<T> extends Stream<T> {
     if (_controller == null) {
       _createController();
     }
-    _sourceStream = _controller.stream; // Mark stream as set.
-    _controller.close();
+    var controller = _controller!;
+    _sourceStream = controller.stream; // Mark stream as set.
+    controller.close();
```
natebosch: [optional] Could refactor so that `_createController` doesn't  assign the field, but returns it.

Then

```dart
var controller = _controller ??= _createController();
_sourceStream = controller.stream;
controller.close();
```

lib/src/stream_sink_transformer/handler_transformer.dart
```diff
@@ -55,19 +58,24 @@ class _HandlerSink<S, T> implements StreamSink<S> {
...
   @override
-  void addError(error, [StackTrace stackTrace]) {
-    if (_transformer._handleError == null) {
+  void addError(error, [StackTrace? stackTrace]) {
+    var handleError = _transformer._handleError;
+    if (handleError == null) {
       _inner.addError(error, stackTrace);
     } else {
-      _transformer._handleError(error, stackTrace, _safeCloseInner);
+      /// TODO: Update to pass AsyncError.defaultStackTrace(error) once that
```
natebosch: [nit] two `//`

lib/src/cancelable_operation.dart
```diff
@@ -155,15 +158,14 @@ class CancelableCompleter<T> {
...
       : _onCancel = onCancel,
         _inner = Completer<T>() {
-    _operation = CancelableOperation<T>._(this);
+    operation = CancelableOperation<T>._(this);
   }
 
   /// The operation controlled by this completer.
-  CancelableOperation<T> get operation => _operation;
-  CancelableOperation<T> _operation;
+  late final CancelableOperation<T> operation;
```
natebosch: I think this gives the appearance of a setter. There won't be a static error for trying to set it... Is that a bug in the SDK?

jakemac53:
Collecting a list of potential breaking changes here https://github.com/dart-lang/async/issues/109

lib/src/cancelable_operation.dart
jakemac53: https://github.com/dart-lang/async/issues/109

lib/src/cancelable_operation.dart
jakemac53: oh interesting...... cc @leafpetersen

lib/src/result/error.dart
jakemac53: done

lib/src/result/error.dart
jakemac53: added to the issue

lib/src/result/result.dart
jakemac53: done

lib/src/result/result.dart
jakemac53: done

lib/src/stream_completer.dart
jakemac53: done

lib/src/cancelable_operation.dart
jakemac53: Going to leave this for now and follow up if we decide it needs to change.

natebosch: I do think this needs to change. https://github.com/dart-lang/async/pull/139

PR #111, opened 2020-03-20
Drop defaulting for null StackTrace

natebosch:
It is now statically non-null.


lib/src/stream_sink_transformer/handler_transformer.dart
```diff
@@ -88,8 +88,8 @@ class _HandlerSink<S, T> implements StreamSink<S> {
             // stack trace arg non-nullable.
```
jakemac53: nit: remove this todo

jakemac53: actually now that I have read it lol we should just pass a tearoff of `_transformer._handleError` if we can now

PR #112, opened 2020-03-23
--null-safety fixes, make valueOrCancellation return a T?

jakemac53:
This did result in one api change, which I think is for the better anyways. `CancelableOperation.valueOrCancellation` now has a return type of `T?`, where it would have previously thrown at runtime if you didn't provide a cancellation value.

Also added a workaround for https://github.com/dart-lang/sdk/issues/41156 in one of the tests.

### DIFF
--- a/test/stream_queue_test.dart
+++ b/test/stream_queue_test.dart
@@ -1046,6 +1046,9 @@ void main() {
       expect(events.withTransaction(expectAsync1((queue) async {
         expect(await queue.next, 2);
         throw 'oh no';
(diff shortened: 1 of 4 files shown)


test/stream_queue_test.dart
```diff
@@ -1046,6 +1046,9 @@ void main() {
       expect(events.withTransaction(expectAsync1((queue) async {
         expect(await queue.next, 2);
         throw 'oh no';
+        // TODO: Remove after https://github.com/dart-lang/sdk/issues/41156
```
natebosch: Can you mention this in the commit message too? That will make it link in github.

PR #119, opened 2020-06-29
Make StreamQueue start listening immediately to broadcast streams.

lrhn:
When a `StreamQueue` is created for a broadcast stream, it didn't use to listen to the stream immediately, which means that any events sent before the first call to `next` (or any other queue request) would be silently lost. That makes an initial `peek` operation change the behavior of the queue, which is confusing to users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.2
+
+* `StreamQueue` starts listening immediately to broadcast strings.
+
 ## 2.4.1
 
 * Deprecate `DelegatingStream.typed`. Use `Stream.cast` instead.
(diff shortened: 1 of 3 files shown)


natebosch:
So far I've only found one use case that this breaks (a test was mocking a `Stream` and didn't implement `isBroadcast`).

I think this is the right thing to do. I do wonder whether we should roll this out before or after NNBD and whether we should model it as a breaking change.

My gut reaction is to treat this as non-breaking and try to get it out ASAP. @jakemac53 - do you have any thoughts?

I'll try to run more tests and see if anything more scare comes up.

jakemac53:
Ya, I would say run it through an internal presubmit and as long as that seems reasonable to land then :shipit:

PR #120, opened 2020-07-01
Remove now unnecessary specified generics

natebosch:
These can now be correctly inferred after a fix in the SDK.

https://github.com/dart-lang/sdk/issues/41106


jakemac53:
Do we need to update the min sdk constraint?

natebosch:
> Do we need to update the min sdk constraint?

I suppose there isn't any reason not to.

PR #123, opened 2020-07-07
fix: StreamQueue handles mock Stream

### LINKED ISSUE #122: breaking change in v2.4.2
## Problem

When creating mock streams using [mockito](https://pub.dev/packages/mockito) before v2.4.2 it was possible to create mocks like:

```dart
import 'package:mockito/mockito.dart';

class MockStream extends Mock implements Stream {}
```

The following would then work:

```dart
final stream = MockStream();
when(stream.listen(
    any,
    onError: captureAnyNamed('onError'),
    onDone: captureAnyNamed('onDone'),
    cancelOnError: captureAnyNamed('cancelOnError'),
  )).thenAnswer((invocation) {
    return Stream.value(0).listen(
      invocation.positionalArguments.first as Function(int),
      onError: invocation.namedArguments[const Symbol('onError')] as Function,
      onDone:invocation.namedArguments[const Symbol('onDone')] as void Function(),
      cancelOnError:invocation.namedArguments[const Symbol('cancelOnError')] as bool,
    );
});
```

After v2.4.2 the above code throws an exception because `_source.isBroadcast` returns `null`.
https://github.com/dart-lang/async/blob/e0c41e0c1fb70ba606dc81335c981af7c00987c6/lib/src/stream_queue.dart#L123

In order to make the above code work we now must additionally stub `isBroadcast`

```dart
final stream = MockStream();
when(stream.isBroadcast).thenReturn(false);
when(stream.listen(...));
```

## Proposal

Add a null check to the existing code for backward compatibility:

```dart
// `isBroadcast` can be null for a mock source.
if (_source?.isBroadcast == true) {
  _ensureListening();
  _pause();
}
```

felangel:
closes #122

### DIFF
--- a/lib/src/stream_queue.dart
+++ b/lib/src/stream_queue.dart
@@ -122,7 +122,7 @@ class StreamQueue<T> {
     // Start listening immediately if we could otherwise lose events.
     // isBroadcast can be null when using [mockito](https://pub.dev/packages/mockito)
     // to create a mock stream.
-    if (_source?.isBroadcast == true) {
+    if (_source.isBroadcast ?? false) {
       _ensureListening();
       _pause();
     }
--- a/lib/src/stream_queue.dart
+++ b/lib/src/stream_queue.dart
@@ -120,7 +120,9 @@ class StreamQueue<T> {
   // Private generative constructor to avoid subclasses.
   StreamQueue._(this._source) {
     // Start listening immediately if we could otherwise lose events.
-    if (_source.isBroadcast) {
+    // isBroadcast can be null when using [mockito](https://pub.dev/packages/mockito)
+    // to create a mock stream.
+    if (_source?.isBroadcast == true) {
       _ensureListening();
       _pause();
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.3
+
+* `StreamQueue` handles mock `Stream` for backward compatibility.
+
 ## 2.4.2
 
 * `StreamQueue` starts listening immediately to broadcast strings.
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: async
-version: 2.4.2
+version: 2.4.3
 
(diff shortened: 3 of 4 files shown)


lib/src/stream_queue.dart
```diff
@@ -120,7 +120,9 @@ class StreamQueue<T> {
   // Private generative constructor to avoid subclasses.
   StreamQueue._(this._source) {
     // Start listening immediately if we could otherwise lose events.
-    if (_source.isBroadcast) {
+    // isBroadcast can be null when using [mockito](https://pub.dev/packages/mockito)
+    // to create a mock stream.
+    if (_source?.isBroadcast == true) {
```
natebosch, line 125: The `?.` isn't necessary here. `_source` should always be non-null.

```suggestion
    if (_source.isBroadcast ?? false) {
```

That said, this feels a little sketchy to me. After NNBD the analyzer won't even want us to write it like this.

It's definitely unfortunate that this surfaces as broken tests, but I do think that the better solution is likely to make the mocks compatible rather than work around it.

@jakemac53 @lrhn - what do you think?

felangel: Yeah I was just trying to be as safe as possible but you're right it shouldn't be necessary in this case.
I feel it would be best to be backward compatible until NNBD lands in stable then this code could be removed with the migration.

natebosch:
After discussing this I don't think it will make sense to try to work around the change and then roll that back along with null safety.

In this case tests that are impacted will need to be updated. In general I think it's safest to avoid using `Mock` with `Stream` - generally prefer to use `StreamController`.

lrhn:
I agree that fixing this in the platform libraries is the wrong place. 

Mocks cab be fragile, here shown by one not implementing an entire API and a test failing when an implementation changes. This change is actually visible, at least if `isBroadcast` returns true, but similar issues could happen for completely invisible changes in implementation details. Say, if you only implement one of `length`, `isEmpty` or `isNotEmpty` of an iterable/list, and an implementation switched between them. 

Partial implementations are not something we promise not to break.

PR #127, opened 2020-07-15
Fix error in CancelableOperation.then

whesse:
Replace a dynamic null check with a dynamic type cast, since the target type may be nullable.


whesse:
I don't know if you also want to change the package version in pubspec.yaml

jakemac53:
The pubspec version does not need to change - it hasn't yet been published (this is why we wait until landing in google3 before publishing!)

PR #132, opened 2020-07-25
Invalidate async cache on exception

himamis:
When the callback throws an exception, the startStaleTime was not called, so the cache was not refreshed. Because of this, future calls to this cache always returned with the exception.
This pull request solve this issue, by wrapping the `await future` call in a try/finally block.

### DIFF
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -60,15 +60,12 @@ class AsyncCache<T> {
     if (_cachedStreamSplitter != null) {
       throw StateError('Previously used to cache via `fetchStream`');
     }
-    if (_cachedValueFuture == null) {
-      _cachedValueFuture = callback();
-      try {
-        await _cachedValueFuture;
-      } finally {
-        _startStaleTimer();
-      }
+    final result = _cachedValueFuture ??= callback();
+    try {
+      return await result;
+    } finally {
+      _startStaleTimer();
     }
-    return _cachedValueFuture!;
   }
 
   /// Returns a cached stream from a previous call to [fetchStream], or runs
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -62,8 +62,11 @@ class AsyncCache<T> {
     }
     if (_cachedValueFuture == null) {
       _cachedValueFuture = callback();
-      await _cachedValueFuture;
-      _startStaleTimer();
+      try {
+        await _cachedValueFuture;
+      } finally {
+        _startStaleTimer();
+      }
     }
     return _cachedValueFuture!;
   }
--- a/test/async_cache_test.dart
+++ b/test/async_cache_test.dart
@@ -158,4 +158,20 @@ void main() {
     }));
     expect(cache.fetchStream(call).toList(), completion(['1', '2', '3']));
   });
+
+  test('should invalidate even if the future throws an exception', () async {
+    cache = AsyncCache.ephemeral();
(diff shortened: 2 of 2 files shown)


himamis:
@googlebot I signed it!

himamis:
* Note: Tests are missing for this change.

kevmoo:
> * Note: Tests are missing for this change.

A test would be really nice before we land this!

himamis:
@kevmoo I have added a unit test.

mit-mit:
@natebosch @lrhn either of you reviewing this?

lib/src/async_cache.dart
```diff
@@ -62,8 +62,11 @@ class AsyncCache<T> {
...
       _cachedValueFuture = callback();
-      await _cachedValueFuture;
-      _startStaleTimer();
+      try {
+        await _cachedValueFuture;
+      } finally {
+        _startStaleTimer();
+      }
     }
     return _cachedValueFuture!;
```
lrhn, line 71: The code here looks slightly problematic. It returns a future *after* awaiting that same future, and reads and null-checks `_cachedValueFuture` repeatedly.

I'd write everything from line 63 as:

```dart
  var result = _cachedValueFuture ??= callback();
  try {
    return await result;
  } finally {
    _startStaleTimer();
  }
```

This should not change the behavior of the code, but it also doesn't need to repeatedly read `_cachedValueFuture`, and it doesn't have to await it twice (once in the `await`, and once in the `return`).

test/async_cache_test.dart
```diff
@@ -158,4 +158,20 @@ void main() {
...
+
+  test('should invalidate even if the future throws an exception', () async {
+    cache = AsyncCache.ephemeral();
+
+    Future<String> throwingCall() async => throw Exception();
+    try {
+      await cache.fetch(throwingCall);
+    } catch (exception) {
+      expect(exception, isException);
+    }
```
lrhn, line 170: Use:
```dart
 await expectLater(cache.fetch(throwingCall), throwsA(isException));
```
so that you check that it actually throws. The code above will succeed even if it *doesn't* throw.

test/async_cache_test.dart
```diff
@@ -158,4 +158,20 @@ void main() {
...
+    // To let the timer invalidate the cache
+    await Future.delayed(Duration.zero);
```
lrhn, line 172: This is probably safe. I'd consider using a non-null duration, just to be absolutely sure that it triggers later than the other zero-duration timer, even if they are scheduled within the same millisecond.
(Just to be safe, probably not an issue in practice).

lib/src/async_cache.dart
himamis: Thanks, it's much better this way.

test/async_cache_test.dart
himamis: I've changed the code to the suggested one, it works well.

test/async_cache_test.dart
himamis: I've added 5 milliseconds to the duration.

mit-mit:
@lrhn looks like this is ready for another review

natebosch (approved):

natebosch:
Looks like this could use a changelog.

@lrhn - if you think the code looks good I can add a changelog and get this landed.

lrhn (commented):
LGTM! Thank you for the fix to this subtle issue.

PR #133, opened 2020-08-12
Update to published package versions

natebosch:
- Remove most dependency overrides. Retain `fake_async` which has not
  been published with a null safe version.
- Update `dev_dependencies` to reference the min null safe versions.
- Remove trailing whitespace.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -16,7 +16,3 @@ dev_dependencies:
   stack_trace: ^1.10.0-nullsafety
   test: ^1.16.0-nullsafety
   pedantic: ^1.10.0-nullsafety
-
-dependency_overrides:
-  fake_async:
-    git: git://github.com/dart-lang/fake_async.git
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -11,72 +11,12 @@ environment:
...
+  test: ^1.16.0-nullsafety
+  pedantic: ^1.10.0-nullsafety
 
 dependency_overrides:
-  # Dev dep overrides
-  clock:
-    git:
-      url: git://github.com/dart-lang/clock.git
-      ref: null_safety
   fake_async:
```
jakemac53, line 21: fake_async has also been published so this should be removable

natebosch: Hmm, `fake_async` has both `1.1.0` and `1.1.0-nullsafety` published.

natebosch: https://github.com/dart-lang/fake_async/pull/23

mit-mit:
Do we have a resolution?

natebosch:
We can punt for now and revisit after we have 2.11 SDKs and have worked out that transition.

PR #135, opened 2020-08-24
Resolve null safety migration TODOS

natebosch:
- Switch from `StackTrace.fromString` to `AsyncError.defaultStackTrace`.
- Remove a TODO about non-nullable `StackTrace`, by the time we merged
  null safety the type was already non-nullable and the TODO was stale.

Style tweak - remove the argument names from a Function type that are
redundant against their type names.

### DIFF
--- a/lib/src/stream_sink_transformer/handler_transformer.dart
+++ b/lib/src/stream_sink_transformer/handler_transformer.dart
@@ -11,11 +11,7 @@ import '../delegate/stream_sink.dart';
 typedef HandleData<S, T> = void Function(S data, EventSink<T> sink);
 
 /// The type of the callback for handling error events.
-//
-// TODO: Update to take a non-nullable StackTrace once that change lands in
-// the sdk.
-typedef HandleError<T> = void Function(
-    Object error, StackTrace stackTrace, EventSink<T> sink);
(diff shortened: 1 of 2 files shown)


lib/src/stream_sink_transformer/handler_transformer.dart
```diff
@@ -11,11 +11,7 @@ import '../delegate/stream_sink.dart';
 typedef HandleData<S, T> = void Function(S data, EventSink<T> sink);
 
 /// The type of the callback for handling error events.
-//
-// TODO: Update to take a non-nullable StackTrace once that change lands in
-// the sdk.
-typedef HandleError<T> = void Function(
-    Object error, StackTrace stackTrace, EventSink<T> sink);
+typedef HandleError<T> = void Function(Object error, StackTrace, EventSink<T>);
```
jakemac53, line 14: it feels weird to have the name specified for one of these and not the others, but 🤷‍♂️

natebosch: It's the only one where the type doesn't tell you what it is.

jakemac53: Ya, I get that, but it still feels weird to me. I would be fine with removing it, as the name of the typedef itself I think makes it pretty clear, or I would also be fine with it as you have it.

PR #150, opened 2020-12-25
Add package of the week video

tvolkert:


tvolkert:
@natebosch

natebosch:
The video doesn't give much information. The video seems like a really good way to help more folks discover the package. After they know it exists and are looking at the docs I don't know what value they would get from a link to the video.

tvolkert:
The package is listed in api.flutter.dev, so they may just click on it to see what it has to offer without any context.  For those users, I think the video still provides value.

natebosch:
cc @lrhn WDYT?

tvolkert:
fwiw, here's what users see right now:

async

lib/async.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// {@youtube 560 315 https://www.youtube.com/watch?v=r0tHiCjW2w0}
+
```
lrhn, line 6: This is a doc comment, so you should probably add a `library` declaration to make it document the library.

We should also add a real library documentation, but for now, let's just have a headline before the link:
```dart
/// Various functionality associated with the asynchronous features of the `dart:async` library.
///
/// {@youtube 560 315 https://www.youtube.com/watch?v=r0tHiCjW2w0}
library pkg.async;
```

lrhn (approved):

lib/async.dart
natebosch: > you should probably add a `library` declaration to make it document the library.

Dartdoc should handle it either way - but I'm in favor of the explicit library declaration for now.

I _really_ hope we can make that `library;` soon.

If I can nitpick the wording, I think "Various" and "functionality" aren't conveying much - virtually any package can be describe as having "Various functionality". How about:

```dart
/// Utilities that expand on the asynchronous features of the `dart:async` library.
///
/// {@youtube 560 315 https://www.youtube.com/watch?v=r0tHiCjW2w0}
library pkg.async;
```

natebosch (approved):
Adding the doc and link LGTM once it has a header in addition to the link.

lib/async.dart
lrhn: That would LGTM

tvolkert:
Ok, all done.  @natebosch can you merge for me?

PR #154, opened 2021-03-03
Use onError over catchError

natebosch:
`Future.catchError` is more error prone in null safe dart than `onError`
because return type of the `Function` callback is not statically
enforced and `null` was a common default value which no longer works.

Migrate all calls from `catchError` to `onError` to demonstrate best
practices. The other usages in this package are mostly on
`Future<dynamic>` and mostly look safe, though I can't rule out a type
error in unusual situations.

There was at least one mistake made in the migration of this library -
the code to ignore errors when completing a `CancelableOperation` as an
error after it was already canceled would instead cause a type error to
surface. Fix this and add a test.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -189,7 +189,7 @@ class CancelableCompleter<T> {
     final future = value as Future<T>;
     if (_isCanceled) {
       // Make sure errors from [value] aren't top-leveled.
-      future.catchError((_) {});
+      future..then<void>((_) {}).onError((_, __) {});
       return;
     }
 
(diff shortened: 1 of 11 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -189,7 +189,7 @@ class CancelableCompleter<T> {
     final future = value as Future<T>;
     if (_isCanceled) {
       // Make sure errors from [value] aren't top-leveled.
-      future.catchError((_) {});
+      future..then<void>((_) {}).onError((_, __) {});
```
lrhn, line 192: I'd do: `future.then<void>((_) {}, onError: (_, __) {})` here. It's guaranteed type-safe because of the `void`, so no need to create two futures where one will do.

lrhn (approved):

lib/src/cancelable_operation.dart
lrhn: The SDK now has an `.ignore()` extension method on futures which ignores errors and does not create a new future that you need to handle.

PR #161, opened 2021-03-29
Add ChunkedStreamReader

jonasfj:
This adds a `ChunkedStreamReader` as a replacement for [`ChunkedStreamIterator` from `package:chunked_stream`](https://github.com/google/dart-neats/blob/4f9e2ce596c71925a4de26cce96b3193d949da3e/chunked_stream/lib/src/chunked_stream_iterator.dart).

Motivation: We would like to use `ChunkedStreamIterator` in Dart SDK and avoid dependency on `package:chunked_stream`. Also for consuming a chunked byte stream in pull-fashion, this has proven very useful :)

I refactored the code a bit, mostly just cleanup, better documentation comments, and rename the class so that code that imports both `package:async` and `package:chunked_stream` won't get a conflict. Test cases are the same to ensure easy migration.

------

Credits @walnutdust for the original `ChunkedStreamIterator` in `package:chunked_stream`.

### DIFF
--- a/lib/src/chunked_stream_reader.dart
+++ b/lib/src/chunked_stream_reader.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:collection';
 import 'dart:typed_data';
 
 import 'byte_collector.dart' show collectBytes;
@@ -35,11 +36,11 @@ import 'byte_collector.dart' show collectBytes;
 /// }
 /// ```
 ///
-/// The read-operations: [readChunk] and [readStream] cannot be invoked
-/// concurrently.
+/// The read-operations [readChunk] and [readStream] must not be invoked until
+/// the future from a previous call has completed.
 class ChunkedStreamReader<T> {
   final StreamIterator<List<T>> _input;
-  final List<T> _emptyList = <T>[];
+  final List<T> _emptyList = UnmodifiableListView([]);
   List<T> _buffer = <T>[];
   bool _reading = false;
 
@@ -55,8 +56,8 @@ class ChunkedStreamReader<T> {
   /// elements have been buffered, or end-of-stream, then it returns the first
   /// [size] buffered elements.
   ///
-  /// If end-of-stream is encountered before [size] bytes is read, this returns
-  /// a list less than [size] (indicating end-of-stream).
+  /// If end-of-stream is encountered before [size] elements is read, this
+  /// returns a list with fewer than [size] elements (indicating end-of-stream).
   ///
   /// If the underlying stream throws, the stream is cancelled, the exception is
   /// propogated and further read operations will fail.
@@ -75,8 +76,8 @@ class ChunkedStreamReader<T> {
   /// This will pass-through _chunks_ from the underlying _chunked stream_ until
   /// [size] elements have been returned, or end-of-stream has been encountered.
   ///
-  /// If end-of-stream is encountered before [size] bytes is read, this returns
-  /// a _chunked stream_ with less than [size] (indicating end-of-stream).
+  /// If end-of-stream is encountered before [size] elements is read, this
+  /// returns a list with fewer than [size] elements (indicating end-of-stream).
   ///
   /// If the underlying stream throws, the stream is cancelled, the exception is
   /// propogated and further read operations will fail.
@@ -87,17 +88,13 @@ class ChunkedStreamReader<T> {
   ///
   /// Throws, if another read-operation is on-going.
   Stream<List<T>> readStream(int size) {
-    if (size < 0) {
-      throw ArgumentError.value(size, 'size', 'must be non-negative');
-    }
+    RangeError.checkNotNegative(size, 'size');
     if (_reading) {
       throw StateError('Concurrent read operations are not allowed!');
     }
     _reading = true;
 
-    final c = StreamController<List<T>>();
-
-    c.addStream(() async* {
+    final substream = () async* {
       // While we have data to read
       while (size > 0) {
         // Read something into the buffer, if it's empty
@@ -111,30 +108,31 @@ class ChunkedStreamReader<T> {
           _buffer = _input.current;
         }
 
-        if (size < _buffer.length) {
-          final output = _buffer.sublist(0, size);
-          _buffer = _buffer.sublist(size);
-          size = 0;
-          yield output;
-          _reading = false;
-          break;
-        }
-
         if (_buffer.isNotEmpty) {
+          if (size < _buffer.length) {
+            final output = _buffer.sublist(0, size);
+            _buffer = _buffer.sublist(size);
+            size = 0;
+            yield output;
+            _reading = false;
+            break;
+          }
+
           final output = _buffer;
           size -= _buffer.length;
           _buffer = _emptyList;
           yield output;
         }
       }
-    }()).whenComplete(c.close);
+    };
 
+    final c = StreamController<List<T>>();
+    c.onListen = () => c.addStream(substream()).whenComplete(c.close);
     c.onCancel = () async {
       while (size > 0) {
         if (_buffer.isEmpty) {
           if (!await _input.moveNext()) {
             size = 0; // no more data
-            _reading = false;
             break;
           }
           _buffer = _input.current;
@@ -143,13 +141,13 @@ class ChunkedStreamReader<T> {
         if (size < _buffer.length) {
           _buffer = _buffer.sublist(size);
           size = 0;
-          _reading = false;
           break;
         }
 
         size -= _buffer.length;
         _buffer = _emptyList;
       }
+      _reading = false;
     };
 
     return c.stream;
@@ -157,7 +155,11 @@ class ChunkedStreamReader<T> {
 
   /// Cancel the underlying _chunked stream_.
   ///
-  /// It is always safe to call [cancel], even if the undelying stream was read
+  /// If a future from [readChunk] or [readStream] is still pending then
+  /// [cancel] behaves as if the underlying stream ended early. That is a future
+  /// from [readChunk] may return a partial chunk smaller than the request size.
+  ///
+  /// It is always safe to call [cancel], even if the underlying stream was read
   /// to completion.
   ///
   /// It can be a good idea to call [cancel] in a `finally`-block when done
--- a/lib/src/chunked_stream_reader.dart
+++ b/lib/src/chunked_stream_reader.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:collection';
 import 'dart:typed_data';
 
 import 'byte_collector.dart' show collectBytes;
@@ -40,7 +39,7 @@ import 'byte_collector.dart' show collectBytes;
 /// the future from a previous call has completed.
 class ChunkedStreamReader<T> {
   final StreamIterator<List<T>> _input;
-  final List<T> _emptyList = UnmodifiableListView([]);
+  final List<T> _emptyList = const [];
   List<T> _buffer = <T>[];
   bool _reading = false;
 
--- a/lib/src/chunked_stream_reader.dart
+++ b/lib/src/chunked_stream_reader.dart
@@ -1,0 +1,176 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:typed_data';
+
+import 'byte_collector.dart' show collectBytes;
+
+/// Utility class for reading elements from a _chunked stream_.
+///
+/// A _chunked stream_ is a stream where each event is a chunk of elements.
+/// Byte-streams with the type `Stream<List<int>>` is common of example of this.
+/// As illustrated in the example below, this utility class makes it easy to
+/// read a _chunked stream_ using custom chunk sizes and sub-stream sizes,
+/// without managing partially read chunks.
+///
+/// ```dart
+/// final r = ChunkedStreamReader(File('myfile.txt').openRead());
+/// try {
+///   // Read the first 4 bytes
+///   final firstBytes = await r.readChunk(4);
+///   if (firstBytes.length < 4) {
+///     throw Exception('myfile.txt has less than 4 bytes');
+///   }
+///
+///   // Read next 8 kilobytes as a substream
+///   Stream<List<int>> substream = r.readStream(8 * 1024);
+///
+///   ...
+/// } finally {
+///   // We always cancel the ChunkedStreamReader, this ensures the underlying
+///   // stream is cancelled.
+///   r.cancel();
+/// }
+/// ```
+///
+/// The read-operations: [readChunk] and [readStream] cannot be invoked
+/// concurrently.
+class ChunkedStreamReader<T> {
+  final StreamIterator<List<T>> _input;
+  final List<T> _emptyList = <T>[];
+  List<T> _buffer = <T>[];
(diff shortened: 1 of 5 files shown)


jonasfj:
Note: I opted to not have `ChunkStreamReader` take a `makeList` in the constructor, but instead have `ChunkedStreamReaderByteStreamExt`. Because it's mostly useful for byte-streams, and, when using it you really want a  `Uint8List` and not a `List<int>` as the return value.

natebosch:
Can you share any of the real world use cases you have in mind for this?

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+///
+///   ...
+/// } finally {
+///   // We always cancel the ChunkedStreamReader, this ensures the underlying
+///   // stream is cancelled.
+///   r.cancel();
+/// }
+/// ```
+///
+/// The read-operations: [readChunk] and [readStream] cannot be invoked
```
lrhn, line 38: Drop the colon.
"cannot" -> "must not"

Consider elaborating on what "concurrently" means: Do not call either method until the future returned by a previous call has completed.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+/// concurrently.
+class ChunkedStreamReader<T> {
+  final StreamIterator<List<T>> _input;
+  final List<T> _emptyList = <T>[];
```
lrhn, line 42: This list is growable. If you share it with someone else, it might not stay empty. I don't *think* you do, it's just used as a sentinel. Why not just use `null` instead to represent the empty buffer, if you are going to check for whether it's empty before using it anyway?
You're going to replace the empty _buffer with something non-empty anyway, and checking `_buffer == null` vs `_buffer.isEmpty` seems like the former should be more efficient.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+  /// If the underlying stream throws, the stream is cancelled, the exception is
+  /// propogated and further read operations will fail.
+  ///
+  /// If the sub-stream returned from [readStream] is cancelled the remaining
+  /// unread elements up-to [size] are drained, allowing subsequent
+  /// read-operations to proceed after cancellation.
+  ///
+  /// Throws, if another read-operation is on-going.
+  Stream<List<T>> readStream(int size) {
+    if (size < 0) {
```
lrhn, line 90: Can use:
     RangeError.checkNotNegative(size, "size");

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+      throw ArgumentError.value(size, 'size', 'must be non-negative');
+    }
+    if (_reading) {
+      throw StateError('Concurrent read operations are not allowed!');
+    }
+    _reading = true;
+
+    final c = StreamController<List<T>>();
+
+    c.addStream(() async* {
```
lrhn, line 100: I'd love to try to optimize this. 
It seems OK for size, but the `async*` function could probably just be an `async` function where `yield e` is replaced by `c.add(e)`. Don't worry about it for now if it works.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+        if (size < _buffer.length) {
+          final output = _buffer.sublist(0, size);
+          _buffer = _buffer.sublist(size);
+          size = 0;
+          yield output;
+          _reading = false;
+          break;
+        }
+
+        if (_buffer.isNotEmpty) {
```
lrhn, line 123: This line can be moved above the `if (size < _buffer.length) {`.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+          final output = _buffer;
+          size -= _buffer.length;
+          _buffer = _emptyList;
+          yield output;
+        }
+      }
+    }()).whenComplete(c.close);
+
+    c.onCancel = () async {
+      while (size > 0) {
```
lrhn, line 133: So, the `addStream` is cancelled first, cancelling the `async*` function above. That happens at a `yield`. I does look like that leaves it at a consistent state. Good.
The `_reading` might be true. Probably definitely will be. It's possible to have `size` be zero if the last chunk was exactly the right size (exit at the yield in line 127 with size == 0), or just breaking at the yield in line 118. In that case `_reading` stays true. You might want a `_reading = false;` at the end of `onCancel`.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /// Cancel the underlying _chunked stream_.
+  ///
+  /// It is always safe to call [cancel], even if the undelying stream was read
+  /// to completion.
+  ///
+  /// It can be a good idea to call [cancel] in a `finally`-block when done
+  /// using the [ChunkedStreamReader], this mitigates risk of leaking resources.
+  Future<void> cancel() async => await _input.cancel();
```
lrhn, line 165: Should there be a state check here? No cancelling while a read is in progress, otherwise ... well, either you get a state error from the stream iterator immediately, or if you do this at a yield-point, the next `moveNext` will return false. That seems safe, but possibly surprising that the input is cut short from the side.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/// Extensions for using [ChunkedStreamReader] with byte-streams.
+extension ChunkedStreamReaderByteStreamExt on ChunkedStreamReader<int> {
+  /// Read bytes into a [Uint8List].
+  ///
+  /// This does the same as [readChunk], except it uses [collectBytes] to create
+  /// a [Uint8List], which offers better performance.
```
lrhn, line 173: And because you only use `sublist` internally, which preserves typed_data lists, it's going to be `Uint8List` all the way down. Nice!

test/chunked_stream_reader.dart
```diff
@@ -0,0 +1,360 @@
...
+      [1, 2, 3],
+      [4],
+    ]));
+    expect(await s.readBytes(1), equals([1]));
+    expect(await s.readBytes(1), isA<Uint8List>());
+    expect(await s.readBytes(1), equals([3]));
+    expect(await s.readBytes(1), equals([4]));
+    expect(await s.readBytes(1), equals([]));
+  });
+}
```
lrhn, line 360: Maybe do some tests with chunk size, and doing operations ending at chunk boundaries and not at chunk boundaries. 
And check cancelling a readStream right after the final chunk, at a chunk boundary, and see if I'm right it won't set `_reading` to false.

lib/src/chunked_stream_reader.dart
lrhn, line 100: The `addStream` immediately listens to the stream, then pauses it if no-one is listening.

That means the internal code will run until the first yield, whether anyone listens to the result of `readStream` or not.
Consider only doing the `addStream` in the `onListen` of `c`.

lib/src/chunked_stream_reader.dart
```diff
@@ -0,0 +1,176 @@
...
+    return result;
+  }
+
+  /// Read next [size] elements from _chunked stream_ as a sub-stream.
+  ///
+  /// This will pass-through _chunks_ from the underlying _chunked stream_ until
+  /// [size] elements have been returned, or end-of-stream has been encountered.
+  ///
+  /// If end-of-stream is encountered before [size] bytes is read, this returns
+  /// a _chunked stream_ with less than [size] (indicating end-of-stream).
```
lrhn, line 79: less -> fewer

lrhn (approved):
Cool stuff!

simolus3:
I tried to integrate this into `package:tar` and ran into some unexpected errors about concurrent reads:

```dart
import 'dart:typed_data';

Future<void> main() async {
  final block = Uint8List(512);
  final stream = Stream.fromIterable(Iterable.generate(10, (_) => block));

  final reader = ChunkedStreamReader(stream);
  while (true) {
    final chunk = await reader.readBytes(1024);

    if (chunk.isEmpty) break;
  }
}
```

This fails with `Bad state: Concurrent read operations are not allowed!`, but it appears to me that there aren't any concurrent reads. Am I overlooking something?

lib/src/chunked_stream_reader.dart
jonasfj: > if you are going to check for whether it's empty before using it anyway?

Well, we also handle the cases when `StreamIterator.current` is an empty list efficiently, without any additional checks.

It's nice to avoid nullable when empty-list should have the same behavior.

How about wrapping with `UnmodifiableListView` from `dart:collection`, that would also guard against mutation of `_emptyList`.

lib/src/chunked_stream_reader.dart
jonasfj: Love it! I didn't know that one! :D

lib/src/chunked_stream_reader.dart
jonasfj: Then we get no back-pressure.. We'd have to also hook `onPause` / `onResume`.

Yes, it could probably be a tiny bit faster, but I'm generally very weary of using `StreamController.add` because it doesn't return a future I can await.

lib/src/chunked_stream_reader.dart
jonasfj: That's a good catch, also fixes the issue @simolus3 hit below.

lib/src/chunked_stream_reader.dart
jonasfj: From reading the `StreamIterator` code I get the impression that `StreamIterator.cancel` makes any pending future from `StreamIterator.moveNext` return `false`.

So if `ChunkedStreamReader.cancel` is called while a future `readChunk` is pending, we get the same behavior as if the underlying stream ended early. Namely `readChunk(size)` may return a chunk smaller than the requested `size`.

I added a test to cover both cases...

lrhn: You can cancel while moving! Yey! Then LGTM.

lib/src/chunked_stream_reader.dart
natebosch: You could make it const. `final List<T> _empty = const [];` or `final List<T> _empty = const <Never>[];`

test/chunked_stream_reader.dart
jonasfj: You are absolute right, bug fixed -- thanks!

nex3:
I know this already landed, but since it hasn't been released yet, something that may be worth considering: could this be re-imagined as a set of extension methods on [`StreamQueue<List<T>>`](https://pub.dev/documentation/async/latest/pkg.async/StreamQueue-class.html)? The `StreamQueue` class already provides the underlying semantics of pull-based access to a stream, with the added benefit of well-defined behavior for calling multiple methods synchronously and nice integration with the `test` package. Consider the following potential API:

```dart
extension ChunkedStreamQueueExtension<T> on StreamQueue<List<T>> {
  Future<List<T>> readChunk(int size);

  Stream<List<T>> readChunkedSubstream(int size);
}
```

I think [`withTransaction`](https://pub.dev/documentation/async/latest/pkg.async/StreamQueue/withTransaction.html) would make it possible to implement the behavior quite tersely as well.

jonasfj:
@nex3, hmm, an extension method would have been nice, but I think I'll need a place to cache a partially consumed chunk.

It might have been possible if `reader.readChunk(5)` was allowed to return more than 5 elements, but that it is not. Which makes it very useful for decoding byte streams like tar where an integer says how long the next file is.

----------

If the stream has the type `Stream<List<int>>` then `reader.readChunk(1)` will return a `List<int>` with length one, even when the underlying `Stream<List<int>>` returns everything in chunks of 4kb or 8kb (as many file systems would).

**Example**
```dart
Stream<List<int>> makeStreamWithChunksOfBytes() async* {
  // Byte chunk 1:
  yield [1],

  // Byte chunk 2:
  yield [2,3,4,5];

  // Byte chunk 3:
  yield [6],
}

final reader = ChunkedStreamReader(makeStreamWithChunksOfBytes());

// Read the first two bytes
expect(await reader.readChunk(2), equals([1,2]));

// At this point "Byte chunk 1" and "Byte chunk 2" has been read from the underlying stream
// and all the elements from "Byte chunk 1" have been return the user.
// BUT: elements 3,4,5 from "Byte chunk 2" are now cached inside [reader]

// Read the next 100 bytes (or in this case until the end of the stream)
expect(await reader.readChunk(100), equals([3,4,5,6]));
```

It's mostly useful when reading byte streams..

-------

I don't see how an extension on `StreamQueue` can cache the unconsumed parts of a chunk. I would need to be able to push something back on-top of the queue. I could use StreamQueue.peak` and modify the top-element, but `Uint8List` doesn't allow that.

nex3:
Yeah, I see what you mean—it's not implementable as an extension method on the class's public API. I think you could do it on the private API, though:

```dart

extension ChunkedStreamQueueExt<T> on StreamQueue<List<T>> {
  Stream<List<T>> readChunkedSubstream(int size) {
    if (size < 0) throw RangeError.range(size, 0, null, 'size');
    if (!_isClosed) {
      var request = _ReadChunkedSubstreamRequest<T>(size);
      _addRequest(request);
      return request.stream;
    }
    throw _failClosed();
  }
}

class _ReadChunkedSubstreamRequest<T> extends _EventRequest<List<T>> {
  Stream<List<T>> get stream => _controller.stream;
  final _controller = StreamController<List<T>>(sync: true);

  final int _size;

  int _consumed = 0;

  _ReadChunkedSubstreamRequest(this._size);

  @override
  bool update(QueueList<Result<List<T>>> events, bool isDone) {
    while (events.isNotEmpty) {
      var event = events.removeFirst();
      if (event.isError) {
        event.addTo(_controller);
        continue;
      }

      var data = event.asValue!.value;
      if (_consumed + data.length < _size) {
        _consumed += data.length;
        _controller.add(data);
      } else {
        if (_consumed + data.length > _size) {
          data = data.sublist(0, _size - _consumed);
          events.addFirst(Result.value(data.sublist(_size - _consumed)));
        }
        _controller.add(data);
        _controller.close();
        return true;
      }
    }

    if (isDone) _controller.close();
    return isDone;
  }
}
```

(Warning: I didn't test this!)

lrhn:
The `StreamQueue` was not built for extensibility, so the internal API isn't particularly user friendly.
It might be worth refactoring it to provide an interface that commands can use to access the queue, and then you can add arbitrary externally defined commands using callbacks.
Or maybe jsut add something like `void nextCommand(bool function(Queue<T>, bool isDone) update);`. The callback is called when it's its time to act, and then every time the underlying queue changes, until it calls `close()`.
(The `_EventRequest` class is a single-method interface, so it should just be a function if exposed 6in a public API)

jonasfj:
> I think you could do it on the private API,

I suppose this might be possible. But I'm not sure it's desirable.

Using the internal API is not ideal, nor the worse. Certainly we could as @lrhn points out make a `nextCommand` method in the public API. This might be a powerful tool for advanced use-cases.

But for most users who want to read a byte-stream block by block, you really don't want other methods available. Those are just footguns.

`ChunkedStreamReader` is intended for users who are reading a byte stream and want the chunk boundaries in the underlying byte stream abstracted away. All the methods on `StreamQueue` operate on chunk boundaries of the underlying byte-stream, this is a different abstraction level -- I don't think mixing them will do users any favors.

If using `ChunkedStreamReader` the public methods on `StreamQueue` do not make sense, mixing these would almost always be a footgun.

nex3:
Yeah, I see what you mean. In that case, maybe it should be a *wrapper* rather than an *extension*, so you can get the nice request-sequencing behavior without having methods at two different abstraction levels. That's something that could probably be added to the existing API after a release, though.

PR #164, opened 2021-04-16
Add StreamGroup.onIdle and StreamGroup.isIdle

nex3:
These match the corresponding APIs on FutureGroup.

### DIFF
--- a/lib/src/stream_group.dart
+++ b/lib/src/stream_group.dart
@@ -41,22 +41,32 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   /// See [_StreamGroupState] for detailed descriptions of each state.
   var _state = _StreamGroupState.dormant;
 
-  /// Whether this group has no active streams.
+  /// Whether this group contains no streams.
+  ///
+  /// A [StreamGroup] is idle when it contains no streams, which is the case for
+  /// a newly created group or one where all added streams have been emitted
+  /// done events (or been [remove]d).
+  ///
+  /// If this is a single-subscription group, then cancelling the subscription
+  /// to [stream] will also remove all streams.
   bool get isIdle => _subscriptions.isEmpty;
 
-  /// A broadcast stream that emits an event whenever the last pending stream in
-  /// this group emits a done event (or is removed).
+  /// A broadcast stream that emits an event whenever this group becomes idle.
+  ///
+  /// A [StreamGroup] is idle when it contains no streams, which is the case for
+  /// a newly created group or one where all added streams have been emitted
+  /// done events (or been [remove]d).
   ///
   /// This stream will close when either:
   ///
-  /// * All streams in this group are done (or removed) *and* [close] has been
-  ///   called, or
+  /// * This group is idle *and* [close] has been called, or
   /// * [stream]'s subscription has been cancelled (if this is a
-  ///   single-subscriber group).
+  ///   single-subscription group).
   ///
-  /// Note that this won't fire until [stream] has been listened to.
+  /// Note that events won't be emitted on this stream until [stream] has been
+  /// listened to.
   Stream<void> get onIdle =>
-      (_onIdleController ??= StreamController.broadcast(sync: true)).stream;
+      (_onIdleController ??= StreamController.broadcast()).stream;
 
   StreamController<Null>? _onIdleController;
 
@@ -158,8 +168,8 @@ class StreamGroup<T> implements Sink<Stream<T>> {
     if (_subscriptions.isEmpty) {
       _onIdleController?.add(null);
       if (_closed) {
-        onIdleController?.close();
-        _controller.close();
+        _onIdleController?.close();
+        scheduleMicrotask(_controller.close);
       }
     }
 
--- a/lib/src/stream_group.dart
+++ b/lib/src/stream_group.dart
@@ -42,8 +42,8 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   /// Whether this group has no active streams.
   bool get isIdle => _subscriptions.isEmpty;
 
-  /// A broadcast stream that emits a `null` event whenever the last pending
-  /// stream in this group emits a done event (or is removed).
+  /// A broadcast stream that emits an event whenever the last pending stream in
+  /// this group emits a done event (or is removed).
   ///
   /// This stream will close when either:
   ///
@@ -53,7 +53,7 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   ///   single-subscriber group).
   ///
   /// Note that this won't fire until [stream] has been listened to.
-  Stream<Null> get onIdle =>
+  Stream<void> get onIdle =>
       (_onIdleController ??= StreamController.broadcast(sync: true)).stream;
 
   StreamController<Null>? _onIdleController;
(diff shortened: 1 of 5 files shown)


lib/src/stream_group.dart
```diff
@@ -39,6 +39,25 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   /// See [_StreamGroupState] for detailed descriptions of each state.
   var _state = _StreamGroupState.dormant;
 
+  /// Whether this group has no active streams.
+  bool get isIdle => _subscriptions.isEmpty;
+
+  /// A broadcast stream that emits a `null` event whenever the last pending
```
natebosch, line 45:
```suggestion
  /// A broadcast stream that emits an event whenever the last pending
```

lib/src/stream_group.dart
```diff
@@ -39,6 +39,25 @@ class StreamGroup<T> implements Sink<Stream<T>> {
...
+  ///
+  /// This stream will close when either:
+  ///
+  /// * All streams in this group are done (or removed) *and* [close] has been
+  ///   called, or
+  /// * [stream]'s subscription has been cancelled (if this is a
+  ///   single-subscriber group).
+  ///
+  /// Note that this won't fire until [stream] has been listened to.
+  Stream<Null> get onIdle =>
```
natebosch, line 56:
```suggestion
  Stream<void> get onIdle =>
```

lib/src/stream_group.dart
```diff
@@ -133,7 +152,16 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   Future? remove(Stream<T> stream) {
     var subscription = _subscriptions.remove(stream);
     var future = subscription == null ? null : subscription.cancel();
-    if (_closed && _subscriptions.isEmpty) _controller.close();
+
+    if (_subscriptions.isEmpty) {
+      var onIdleController = _onIdleController;
+      if (onIdleController != null) onIdleController.add(null);
```
natebosch, line 158:
```suggestion
      _onIdleController?.add(null);
```

lib/src/stream_group.dart
```diff
@@ -133,7 +152,16 @@ class StreamGroup<T> implements Sink<Stream<T>> {
...
+      if (_closed) {
+        if (onIdleController != null) onIdleController.close();
```
natebosch, line 160:
```suggestion
        _onIdleController?.close();
```

test/stream_group_test.dart
```diff
@@ -734,6 +734,107 @@ void regardlessOfType(StreamGroup<String> Function() newStreamGroup) {
...
+      streamGroup.onIdle.listen(expectAsync1((_) {
+        expect(streamClosed, isFalse);
+        idle = true;
+      }), onDone: expectAsync0(() {
+        expect(idle, isTrue);
+        expect(streamClosed, isFalse);
+        onIdleDone = true;
+      }));
+
+      streamGroup.stream.toList().then(expectAsync1((_) {
```
natebosch, line 817: [nit] 
```suggestion
      streamGroup.stream.drain().then(expectAsync1((_) {
```

test/stream_group_test.dart
```diff
@@ -734,6 +734,107 @@ void regardlessOfType(StreamGroup<String> Function() newStreamGroup) {
...
+
+      controller.close();
+      await flushMicrotasks();
+      expect(idle, isTrue);
+      expect(streamGroup.isIdle, isTrue);
+    });
+
+    test('emits an event when the group closes', () async {
+      // It's important that the order of events here stays consistent over
+      // time, since code may rely on it in subtle ways.
```
natebosch, line 803: I don't think we should have any enforced ordering between the stream done and the onIdle done.

cc @lrhn for thoughts on testing behavior that we don't want to make guarantees about.

nex3: I think there's a difference between "publicly making a guarantee that the behavior will remain consistent", and "internally trying to ensure that the behavior will remain consistent". I agree that you shouldn't advertise this behavior, or consider it part of the public API; however, I think it's worth doing what you can to avoid breaking it *accidentally* and thereby risking a bunch of subtle bugs without any benefit.

lib/src/stream_group.dart
```diff
@@ -135,7 +154,15 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   Future? remove(Stream<T> stream) {
     var subscription = _subscriptions.remove(stream);
     var future = subscription == null ? null : subscription.cancel();
-    if (_closed && _subscriptions.isEmpty) _controller.close();
+
+    if (_subscriptions.isEmpty) {
+      _onIdleController?.add(null);
+      if (_closed) {
+        onIdleController?.close();
+        _controller.close();
```
lrhn, line 162: This also sends a done event synchronously to the `_controller` stream in response to a synchronous `remove` call. I'd do `scheduleMicrotask(_controller.close);` here.

test/stream_group_test.dart
lrhn: Honest answer: I'd rather randomize things that people shouldn't depend on, so they can learn to not depend on them. It's a pain to change *anything* timing related because people have written bad code depending on the implementation, not the specification.

In practice, I'm fine testing that the implementation does what we expect the implementation to do, as long as it's *clearly documented* that a change of behavior is fine as long as you also change the test. It's often easier to test the actual behavior than test "any of a number of correct behaviors", you'll likelyend up missing some technically correct behaviors anyway.

lib/src/stream_group.dart
```diff
@@ -180,6 +207,13 @@ class StreamGroup<T> implements Sink<Stream<T>> {
         .toList();
 
     _subscriptions.clear();
+
+    var onIdleController = _onIdleController;
+    if (onIdleController != null && !onIdleController.isClosed) {
+      onIdleController.add(null);
+      onIdleController.close();
```
lrhn, line 214: Sends events synchronously here.

lib/src/stream_group.dart
```diff
@@ -135,7 +154,15 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   Future? remove(Stream<T> stream) {
     var subscription = _subscriptions.remove(stream);
     var future = subscription == null ? null : subscription.cancel();
-    if (_closed && _subscriptions.isEmpty) _controller.close();
+
+    if (_subscriptions.isEmpty) {
+      _onIdleController?.add(null);
```
lrhn, line 159: Sends events synchronously here.

lib/src/stream_group.dart
```diff
@@ -41,6 +41,25 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   /// See [_StreamGroupState] for detailed descriptions of each state.
   var _state = _StreamGroupState.dormant;
 
+  /// Whether this group has no active streams.
```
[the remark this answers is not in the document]
nex3: Done. I'll also update `FutureGroup`'s documentation to match.

test/stream_group_test.dart
nex3: Added more explicit documentation.

lib/src/stream_group.dart
```diff
@@ -41,6 +41,35 @@ class StreamGroup<T> implements Sink<Stream<T>> {
...
+  /// done events (or been [remove]d).
+  ///
+  /// This stream will close when either:
+  ///
+  /// * This group is idle *and* [close] has been called, or
+  /// * [stream]'s subscription has been cancelled (if this is a
+  ///   single-subscription group).
+  ///
+  /// Note that events won't be emitted on this stream until [stream] has been
+  /// listened to.
```
lrhn, line 67: Is it worth mentioning that the events are delivered asynchronously and the group can have become be non-idle when the event is delivered.

lrhn (approved):

lib/src/stream_group.dart
nex3: Good idea, done.

PR #166, opened 2021-04-20
Add a StreamCloser transformer

nex3:
This is essentially the Stream equivalent of the Disconnector class in
stream_channel.

### DIFF
--- a/lib/src/stream_closer.dart
+++ b/lib/src/stream_closer.dart
@@ -1,0 +1,89 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'async_memoizer.dart';
+
+/// A [StreamTransformer] that allows the caller to forcibly close the
+/// transformed [Stream].
+///
+/// When [close] is called, the stream (or streams) transformed by [this] will
+/// emit a done event and cancel the underlying subscription.
+class StreamCloser<T> extends StreamTransformerBase<T, T> {
+  /// Whether [close] has been called.
+  bool get isClosed => _closeMemo.hasRun;
+
+  /// The subscriptions to streams passed to [bind].
+  final _subscriptions = <StreamSubscription<T>>{};
+
+  /// The controllers for streams returned by [bind].
+  final _controllers = <StreamController<T>>{};
+
+  /// Closes all transformed streams.
+  ///
+  /// Returns a future that completes when all inner subscriptions'
+  /// [StreamSubscription.cancel] futures have completed. Note that a stream's
+  /// subscription won't be canceled until the transformed stream has a
+  /// listener.
+  ///
+  /// If a transformed stream is listened to after [close] is called, the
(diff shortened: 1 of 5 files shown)


natebosch:
We'll be merging `package:stream_transform` to this package under a separate library `package:async/stream_transform.dart`. When we do that this utility would have a lot of overlap with `takeUntil`. 
https://pub.dev/documentation/stream_transform/latest/stream_transform/TakeUntil/takeUntil.html

cc @lrhn - do you think there is value in having both in the same package?

nex3:
If we don't want to have both, I think `StreamCloser` is the one to keep, since it avoids the microtask gap between completing `trigger` and canceling the underlying subscription and it provides a mechanism for the caller to handle errors in `StreamSubscription.cancel()`. If we do want to have both, it would be pretty easy to re-implement `takeUntil()` in terms of `StreamCloser`.

lrhn (approved):
I don't see much overlap API-wise between `StreamCloser` and `takeUntil`. I think we can easily have both.

lib/src/stream_closer.dart
```diff
@@ -0,0 +1,106 @@
...
+  /// Returns a future that completes when all inner subscriptions'
+  /// [StreamSubscription.cancel] futures have completed. Note that a stream's
+  /// subscription won't be canceled until the transformed stream has a
+  /// listener.
+  ///
+  /// If a transformed stream is listened to after [close] is called, the
+  /// original stream will be listened to and then the subscription immediately
+  /// canceled. If that cancellation throws an error, it will be silently
+  /// ignored.
+  Future<void> close() => _closeMemo.runOnce(() {
```
lrhn, line 41: Is `AsyncMemoizer` worth the overhead here. It has an extra completer with its future, and a call to `Future.sync`. That's two extra futures and a completer for no real benefit.
Couldn't it just be:
```dart
Future<void>? _closeFuture;
bool get isClosed => _closeFuture != null;
Future<void> close() => _closeFuture ??= _close();
Future<void> _close() {
  // code here.
}
```
(or inline `_close` in `close` as `() { ... } ();`).

lib/src/stream_closer.dart
```diff
@@ -0,0 +1,106 @@
...
+        _controllers.clear();
+        scheduleMicrotask(() {
+          for (var controller in controllers) {
+            scheduleMicrotask(controller.close);
+          }
+        });
+
+        return Future.wait(futures, eagerError: true);
+      });
+  final _closeMemo = AsyncMemoizer();
```
lrhn, line 57: Make it `AsyncMemoizer<void>()` for type correctness.

lib/src/stream_closer.dart
```diff
@@ -0,0 +1,106 @@
...
+
+  @override
+  Stream<T> bind(Stream<T> stream) {
+    var controller = stream.isBroadcast
+        ? StreamController<T>.broadcast(sync: true)
+        : StreamController<T>(sync: true);
+
+    controller.onListen = () {
+      if (isClosed) {
```
lrhn, line 66: Consider using `if (_closeMemo.hasRun) {` here, to be safe against someone extending the class and overriding `isClosed`. Ditto for other uses of the public `isClosed`.

lrhn (approved):

PR #167, opened 2021-04-21
Style tweaks in stream_queue.dart

natebosch:
- Use a noun phrase to document the `hasNext` getter. The `next` and
  `peek` getters keep their verb phrase doc comments since the side
  effects are critical distinctions between these methods. If we were
  writing this today, `next()` would be a method.
- Change `_failClosed()` which returns an error, to `_checkClosed()`
  which checks the condition and optionally closes. Avoid nesting the
  majority of method behavior in a conditional.
- Use `RangeError.checkNotNegative` over a conditional.
- Change bare `Future` to `Future<void>` in code examples.
- Make `withTransaction` `async` since there is no longer a blocker now
  that async methods start running synchronously.

### DIFF
--- a/lib/src/stream_queue.dart
+++ b/lib/src/stream_queue.dart
@@ -126,7 +126,7 @@ class StreamQueue<T> {
     }
   }
 
-  /// Asks if the stream has any more events.
+  /// Whether the stream has any more events.
   ///
   /// Returns a future that completes with `true` if the stream has any
   /// more events, whether data or error.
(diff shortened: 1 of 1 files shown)


lib/src/stream_queue.dart
```diff
@@ -529,12 +516,9 @@ class StreamQueue<T> {
...
   // Internal helper methods.
 
-  /// Returns an error for when a request is made after cancel.
-  ///
-  /// Returns a [StateError] with a message saying that either
-  /// [cancel] or [rest] have already been called.
-  Error _failClosed() {
-    return StateError('Already cancelled');
+  /// Throws an error if [cancel] or [rest] have already been called.
+  void _checkClosed() {
```
jakemac53, line 520: nit: `_checkNotClosed`?

lib/src/stream_queue.dart
```diff
@@ -330,23 +316,24 @@ class StreamQueue<T> {
...
-        transaction.commit(queue);
-      } else {
-        transaction.reject();
-      }
-      return result;
-    }, onError: (Object error) {
+    bool result;
+    try {
+      result = await callback(queue);
+    } on Object {
```
jakemac53, line 327: I think its more common to do `catch (_)` if you want to catch all exceptions and ignore the error, although I don't see a problem with this I think its a bit harder to read. Made me do a double take lol.

lib/src/stream_queue.dart
natebosch: Done

lib/src/stream_queue.dart
natebosch: Done.

jakemac53 (approved):

PR #169, opened 2021-04-27
Add a StreamSink.rejectErrors() extension method

nex3:
This makes it easy for authors to expose sinks that can't natively
consume errors, but still handle them in a consistent and robust
manner.

### DIFF
--- a/lib/src/stream_sink_extensions.dart
+++ b/lib/src/stream_sink_extensions.dart
@@ -13,8 +13,7 @@ extension StreamSinkExtensions<T> on StreamSink<T> {
   StreamSink<S> transform<S>(StreamSinkTransformer<S, T> transformer) =>
       transformer.bind(this);
 
-  /// Returns a [StreamSink] that forwards to [this], but doesn't allow any
-  /// errors to be passed in.
+  /// Returns a [StreamSink] that forwards to [this] but rejects errors.
   ///
   /// If an error is passed (either by [addError] or [addStream]), the
   /// underlying sink will be closed and the error will be forwarded to the
--- a/lib/src/stream_sink_extensions.dart
+++ b/lib/src/stream_sink_extensions.dart
@@ -1,0 +1,23 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'stream_sink_transformer.dart';
+import 'stream_sink_transformer/reject_errors.dart';
+
+/// Extensions on [StreamSink] to make stream transformations more fluent.
+extension StreamSinkExtensions<T> on StreamSink<T> {
+  /// Transforms a [StreamSink] using [transformer].
+  StreamSink<S> transform<S>(StreamSinkTransformer<S, T> transformer) =>
+      transformer.bind(this);
+
+  /// Returns a [StreamSink] that forwards to [this], but doesn't allow any
+  /// errors to be passed in.
+  ///
+  /// If an error is passed (either by [addError] or [addStream]), the
+  /// underlying sink will be closed and the error will be forwarded to the
+  /// returned sink's [StreamSink.done] future. Further events will be ignored.
+  StreamSink<T> rejectErrors() => RejectErrorsSink(this);
+}
--- a/lib/src/stream_sink_transformer/reject_errors.dart
+++ b/lib/src/stream_sink_transformer/reject_errors.dart
@@ -1,0 +1,127 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// A [StreammSink] wrapper that rejects all errors passed into the sink.
+class RejectErrorsSink<T> implements StreamSink<T> {
+  /// The inner sink being wrapped.
+  final StreamSink<T> _inner;
+
+  @override
+  Future<void> get done => _doneCompleter.future;
+  final _doneCompleter = Completer();
+
+  /// Whether the underlying sink is no longer receiving events.
+  ///
+  /// This can happen if:
+  ///
+  /// * [close] has been called,
+  /// * an error has been passed,
+  /// * or the underlying [StreamSink.done] has completed.
+  ///
+  /// If [_canceled] is true, [_inAddStream] must be false.
+  bool get _canceled => _doneCompleter.isCompleted;
+
+  /// Whether the user has called [close].
+  ///
+  /// If [_closed] is true, [_canceled] must be true and [_inAddStream] must be
+  /// false.
+  bool _closed = false;
+
+  /// The subscription to the stream passed to [addStream], if a stream is
+  /// currently being added.
+  StreamSubscription<T>? _addStreamSubscription;
+
+  /// The completer for the future returned by [addStream], if a stream is
+  /// currently being added.
+  Completer<void>? _addStreamCompleter;
+
+  /// Whether we're currently adding a stream with [addStream].
+  bool get _inAddStream => _addStreamSubscription != null;
+
+  RejectErrorsSink(this._inner) {
+    _inner.done.whenComplete(() {
+      if (_inAddStream) {
+        _addStreamCompleter!.complete(_addStreamSubscription!.cancel());
+        _addStreamCompleter = null;
+        _addStreamSubscription = null;
+      }
+    }).then((value) {
+      if (!_canceled) _doneCompleter.complete(value);
+    }).onError<Object>((error, stackTrace) {
+      if (!_canceled) _doneCompleter.completeError(error, stackTrace);
+    });
+  }
+
+  @override
+  void add(T data) {
+    if (_closed) throw StateError('Cannot add event after closing.');
+    if (_inAddStream) {
+      throw StateError('Cannot add event while adding stream.');
+    }
+    if (_canceled) return;
+
+    _inner.add(data);
+  }
+
+  @override
+  void addError(error, [StackTrace? stackTrace]) {
+    if (_closed) throw StateError('Cannot add event after closing.');
+    if (_inAddStream) {
+      throw StateError('Cannot add event while adding stream.');
+    }
+    if (_canceled) return;
+
+    _addError(error, stackTrace);
+  }
+
+  /// Like [addError], but doesn't check to ensure that an error can be added.
+  ///
+  /// This is called from [addStream], so it shouldn't fail if a stream is being
+  /// added.
+  void _addError(Object error, [StackTrace? stackTrace]) {
+    _doneCompleter.completeError(error, stackTrace);
+
+    if (_inAddStream) {
+      _addStreamCompleter!.complete(_addStreamSubscription!.cancel());
+      _addStreamCompleter = null;
+      _addStreamSubscription = null;
+    }
+
+    // Ignore errors from the inner sink. We're already surfacing one error, and
+    // if the user handles it we don't want them to have another top-level.
+    _inner.close().catchError((_) {});
+  }
+
+  @override
+  Future<void> addStream(Stream<T> stream) {
+    if (_closed) throw StateError('Cannot add stream after closing.');
+    if (_inAddStream) {
+      throw StateError('Cannot add stream while adding stream.');
+    }
+    if (_canceled) return Future.value();
+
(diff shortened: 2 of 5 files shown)


lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// A [StreammSink] wrapper that rejects all errors passed into the sink.
```
lrhn, line 7: Typo -> [StreamSink]

lib/src/stream_sink_extensions.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'stream_sink_transformer/reject_errors.dart';
+
+/// Extensions on [StreamSink] to make stream transformations more fluent.
+extension StreamSinkExtensions<T> on StreamSink<T> {
+  /// Transforms a [StreamSink] using [transformer].
+  StreamSink<S> transform<S>(StreamSinkTransformer<S, T> transformer) =>
+      transformer.bind(this);
+
+  /// Returns a [StreamSink] that forwards to [this], but doesn't allow any
+  /// errors to be passed in.
```
lrhn, line 17: First line should be shorter.
The `[this]` is not a valid doc link (not an identifier, not a source declaration).
Generally, never start a dartdoc with "Returns".

Maybe:
> /// Transforms this stream sink into one which rejects errors.
> ///
> /// Returns a new [StreamSink]. Any non-error event added to that new sink,
> /// whether by [add], [close] or [addStream], is also added to this stream sink.
> /// Any error event added to the new sink, whether by [addError] or [addStream],
> /// will instead close this sinks, the error becomes the result of the returned
> /// sink's [StreamSink.done] future, and further events added to the new sink
> /// are ignored.

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+class RejectErrorsSink<T> implements StreamSink<T> {
+  /// The inner sink being wrapped.
```
lrhn, line 9: I'd drop "inner". Maybe just "The target sink".

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+  final StreamSink<T> _inner;
+
+  @override
+  Future<void> get done => _doneCompleter.future;
+  final _doneCompleter = Completer();
```
lrhn, line 14: `Completer<void>();`. No reason to not have the correct type, even if it's unlikely that someone will print `sink.done.runtimeType`.
(I prefer writing the type for fields, even if they're not public, but I guess inference can handle this one).

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+
+import 'dart:async';
+
+/// A [StreammSink] wrapper that rejects all errors passed into the sink.
+class RejectErrorsSink<T> implements StreamSink<T> {
+  /// The inner sink being wrapped.
+  final StreamSink<T> _inner;
+
+  @override
+  Future<void> get done => _doneCompleter.future;
```
lrhn, line 13: I'd move the getters below the constructor.
(State above constructor, members below, makes it easier to figure out the state of the object).

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+  RejectErrorsSink(this._inner) {
+    _inner.done.whenComplete(() {
+      if (_inAddStream) {
+        _addStreamCompleter!.complete(_addStreamSubscription!.cancel());
+        _addStreamCompleter = null;
+        _addStreamSubscription = null;
+      }
+    }).then((value) {
+      if (!_canceled) _doneCompleter.complete(value);
+    }).onError<Object>((error, stackTrace) {
```
lrhn, line 53: While I applaud the use of `onError`, I'd just combine it into one `then` call with an `onError` argument, instead of creating an unnecessary intermediate future.
So, all in all:
```dart
  _inner.done.then((value) {
    _cancelAddStream();
    if (!_canceled) _doneCompleter.complete(value);
  }, onError: (error, stackTrace) {
    _cancelAddStream();
    if (!_canceled) _doneCompleter.completeError(error, stackTrace);
  });
```

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+  /// Whether we're currently adding a stream with [addStream].
+  bool get _inAddStream => _addStreamSubscription != null;
+
+  RejectErrorsSink(this._inner) {
+    _inner.done.whenComplete(() {
+      if (_inAddStream) {
+        _addStreamCompleter!.complete(_addStreamSubscription!.cancel());
+        _addStreamCompleter = null;
+        _addStreamSubscription = null;
+      }
```
lrhn, line 50: Consider making this a helper function and just call it in both the `then` and `onError` branches, instead of creating an extra `Future`.

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+    }).then((value) {
+      if (!_canceled) _doneCompleter.complete(value);
+    }).onError<Object>((error, stackTrace) {
+      if (!_canceled) _doneCompleter.completeError(error, stackTrace);
+    });
+  }
+
+  @override
+  void add(T data) {
+    if (_closed) throw StateError('Cannot add event after closing.');
```
lrhn, line 60: Maybe:
```dart
  if (_canceled) {
    if (_closed) throw StateError(...);
    return;  
  }
```
for one fewer tests on the common path.

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /// Like [addError], but doesn't check to ensure that an error can be added.
+  ///
+  /// This is called from [addStream], so it shouldn't fail if a stream is being
+  /// added.
+  void _addError(Object error, [StackTrace? stackTrace]) {
+    _doneCompleter.completeError(error, stackTrace);
+
+    if (_inAddStream) {
```
lrhn, line 87: Another instance of this code, so using a shared helper function now sound even better.

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+    if (_closed) throw StateError('Cannot add stream after closing.');
+    if (_inAddStream) {
+      throw StateError('Cannot add stream while adding stream.');
+    }
+    if (_canceled) return Future.value();
+
+    var addStreamCompleter = _addStreamCompleter = Completer.sync();
+    _addStreamSubscription = stream.listen(_inner.add,
+        onError: _addError, onDone: addStreamCompleter.complete);
+    return addStreamCompleter.future.then((_) {
```
lrhn, line 109: If the `addStreamCompleter.future` completes with an error then that error will be unhandled here. Add an `onError: (_) {}` to the `then` call as well.
(I think it's fine to not clear the fields in the error branch, the only way that future can complete with an error from one of the places above where the `complete` call is followed by clearing the fields too. Might be worth a comment.)

lib/src/stream_sink_transformer/reject_errors.dart
```diff
@@ -0,0 +1,127 @@
...
+    });
+  }
+
+  @override
+  Future<void> close() {
+    if (_inAddStream) {
+      throw StateError('Cannot close sink while adding stream.');
+    }
+
+    if (_closed) return done;
```
lrhn, line 121: Return `_doneCompleter.future` instead. Just in case someone subclasses the class and overrides the `done` getter, it's better to not depend on the public API in other methods.

lrhn (approved):
Very reasonable functionality.

lib/src/stream_sink_extensions.dart
nex3: > First line should be shorter.

Done.

> The `[this]` is not a valid doc link (not an identifier, not a source declaration).

I don't think this is accurate—the SDK has over 100 instances of `[this]`, and it does seem to be linked by dartdoc (see [this example](https://api.dart.dev/stable/2.12.4/dart-typed_data/Float32x4/scale.html)).

> Generally, never start a dartdoc with "Returns".

That's contrary to [Effective Dart](https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs)'s suggestion.

lib/src/stream_sink_transformer/reject_errors.dart
nex3: This is just the public interface for `_doneCompleter`, which is state. I can move the `canceled` getter below the constructor, although I find it useful to group getters with fields because they both *logically* (if not *physically*) represent state.

lib/src/stream_sink_transformer/reject_errors.dart
nex3: I did that at first, but it doesn't work well with the type system—the `onError` parameter just takes a `Function`, so all the parameters are `dynamic` and would require explicit casts to pass into `Completer.completeError()`.

lib/src/stream_sink_transformer/reject_errors.dart
nex3: I think that's minor enough not to be worth the extra flow complexity.

lib/src/stream_sink_transformer/reject_errors.dart
nex3: This future is returned, so the error should be handled by the caller.

lib/src/stream_sink_transformer/reject_errors.dart
nex3: This class isn't exposed in the package's public API, so there's no way for someone to (legally) subclass it. I think having the terseness here is worth more than the extreme edge case of someone diving into `src` and overriding `done.

lib/src/stream_sink_transformer/reject_errors.dart
lrhn: True, one should type the parameters in that case:
```dart
}, onError: (Objeft error, StackTrace stackTrace) {
    _cancelAddStream();
    if (!_canceled) _doneCompleter.completeError(error, stackTrace);
});
```
No casting needed then.

lib/src/stream_sink_transformer/reject_errors.dart
lrhn: Let's say "internal state" then. When trying to understand a class, as a programmer working on it, I like to have the internal state in one place, and by itself.

I'm aware others do it differently, and the style guide is silent on member ordering.

lib/src/stream_sink_transformer/reject_errors.dart
lrhn: ACK, my bad.

lib/src/stream_sink_transformer/reject_errors.dart
lrhn: ACK

lib/src/stream_sink_transformer/reject_errors.dart
lrhn: Ah, private helper class, that's OK then.
(I really wish Dart had a `self` similar to `super` to do non-virtual self calls, and that unqualified identifiers like `done` defaulted to `self.done` instead of `this.done`. Maybe the next language will be better :).

lib/src/stream_sink_extensions.dart
lrhn: TIL that `[this]` is valid and links to the class itself. I wish DartDoc had, well, documentation (I still don't remember how to link to an unnamed constructor).

I disagree with the Effective Dart *example* here. The suggestion to use active verbs is fine, it's just that the word "Returns" is always inferior to some other word (or none, for getter-like functions where you should use a noun phrase).

If the *only* or *most important* thing you can say about a function is which object it returns, it should be treated as a getter-like function ("parameterized getter") and documented as such (noun phrase).

Most of the time, it's not the most important thing, and the *operation* performed by the function is more relevant. In this case, "Creates" is a better word since it emphasized that the returned sink is *new*.

So, it's not that "Returns" is technically wrong, I've just found that it's just *never* the *best* word to use. Seeing it means that the text can be improved. 


(Having a later section starting with "Returns ..." is highly recommended as the way to describe the returned value. Users should be trained to look for such a section. That too gets muddled when the first overview sentence starts with "Returns" as well).

PR #173, opened 2021-04-28
Fix a StreamGroup bug when a component stream's listen() throws

nex3:
This would put the StreamGroup into an inconsistent state where it
would believe itself to be active, but only some streams would have
subscriptions. This was exacerbated by dart-lang/sdk#45815, which
meant that even though _onListen threw an error a StreamSubscription
was created and returned, so further callbacks could still be called.

Now instead of going into an inconsistent state, the StreamGroup
simply cancels itself.

### DIFF
--- a/lib/src/stream_group.dart
+++ b/lib/src/stream_group.dart
@@ -4,6 +4,8 @@
 
 import 'dart:async';
 
+import 'package:collection/collection.dart';
+
 /// A collection of streams whose events are unified and sent through a central
 /// stream.
 ///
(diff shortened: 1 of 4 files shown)


lib/src/stream_group.dart
```diff
@@ -216,8 +229,17 @@ class StreamGroup<T> implements Sink<Stream<T>> {
   Future? _onCancel() {
     _state = _StreamGroupState.canceled;
 
-    var futures = _subscriptions.values
-        .map((subscription) => subscription!.cancel())
+    var futures = _subscriptions.entries
+        .map((entry) {
+          var subscription = entry.value;
+          if (subscription != null) return subscription.cancel();
```
lrhn, line 235: For total paranoia, consider moving this line inside the `try`. If the `cancel` call throws synchronously (then it's very badly behaved, but) it might cause other subscriptions to not get cancelled.

lrhn (approved):

PR #177, opened 2021-05-07
Don't invoke onValue for a canceled operation

### LINKED ISSUE #172: CancelableOperation executing "then" callback after cancellation
I'm not quite sure if this is a bug or just user error, but I'm a bit confused about the semantics of `CancelableOperation`. Below is a simple example where I create a `CancelableOperation`, register a "then" callback, and then cancel the operation. It will try to execute my "then" callback with a value of `null` (causing a crash). I would have expected my "then" callback not be executed, though perhaps I misunderstand how this is intended to work?

```dart

Future<void> main() async {
  final op = CancelableOperation.fromFuture(
    Future.delayed(const Duration(seconds: 1)).then((_) => "hello, world")
  );
  op.then((value) => print('value: $value'));
  op.cancel();
}
```

Throws the following:

```
Unhandled exception:
type 'Null' is not a subtype of type 'String' in type cast
#0      CancelableOperation.then.<anonymous closure>.<anonymous closure> (package:async/src/cancelable_operation.dart:106:63)
#1      new Future.sync (dart:async/future.dart:218:31)
#2      CancelableOperation.then.<anonymous closure> (package:async/src/cancelable_operation.dart:106:37)
#3      _RootZone.runUnary (dart:async/zone.dart:1612:54)
#4      _FutureListener.handleValue (dart:async/future_impl.dart:152:18)
#5      Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:704:45)
#6      Future._propagateToListeners (dart:async/future_impl.dart:733:32)
#7      Future._complete (dart:async/future_impl.dart:530:7)
#8      _SyncCompleter.complete (dart:async/future_impl.dart:57:12)
#9      CancelableOperation.valueOrCancellation.<anonymous closure> (package:async/src/cancelable_operation.dart:77:17)
#10     _RootZone.runUnary (dart:async/zone.dart:1612:54)
#11     _FutureListener.handleValue (dart:async/future_impl.dart:152:18)
#12     Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:704:45)
#13     Future._propagateToListeners (dart:async/future_impl.dart:733:32)
#14     Future._chainCoreFuture (dart:async/future_impl.dart:507:7)
#15     Future._chainFuture (dart:async/future_impl.dart:590:9)
#16     Future._asyncComplete (dart:async/future_impl.dart:564:7)
#17     _AsyncCompleter.complete (dart:async/future_impl.dart:46:12)
#18     AsyncMemoizer.runOnce (package:async/src/async_memoizer.dart:43:29)
#19     CancelableCompleter._cancel (package:async/src/cancelable_operation.dart:218:24)
#20     CancelableOperation.cancel (package:async/src/cancelable_operation.dart:129:33)
#21     main (file:///test/test.dart:8:6)
#22     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:283:19)
#23     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
```

It appears that after cancelation `CancelableOperation<T>.then` is still executing a "then" callback on the `Future` returned by `valueOrCancellation()`. `valueOrCancellation` returns a `Future<T?>` (nullable I assume because the cancellation value can be `null`). This "then" callback will then cast the value as `T`, even though it is `null`, causing the crash.

This same issue can be triggered with even simpler logic.

```dart
final op = CancelableOperation.fromFuture(Future.value("hello, world"));
op.then((value) => print('value: $value'));
op.cancel();
```

Also, it doesn't seem to matter if the `cancel()` call occurs before or after the `op.then((value) => print('value: $value'))` call.

Explicitly declaring `op` as `CancelableOperation<String?>.fromFuture` will prevent the crash, but it will still execute the then callback with `value` = `null`. To me those semantics seem a bit confusing, though perhaps I'm misunderstanding something?

natebosch:
Fixes #172

The `isCompleted` boolean does not necessarily indicate that
`valueOrCancellation` will complete with a value. If the completer was
completed with a `Future` and then canceled before that argument
invokes the listener callback, the `valueOrCancellation` future will
complete with `null` but `isCompleted` will be true. We only have a
value to pass to `onValue` when the completer was completed and not
canceled.

Add tests for the behavior of `then` when `cancel` is called after
`completer.complete`, both when it was completed with a Future and a
non-Future value.

Move the initialization for `originalCompleter` to the `setup` for the
tests so that it can be used before calling `then`.

### DIFF
--- a/test/cancelable_operation_test.dart
+++ b/test/cancelable_operation_test.dart
@@ -388,7 +388,8 @@ void main() {
         originalCompleter.operation.cancel();
       });
 
-      test('after completing with a future does not invoke `onValue`', () {
+      test('after completing with a future does not invoke `onValue`',
+          () async {
         onValue = expectAsync1((_) => '', count: 0);
         onCancel = null;
         var operation = runThen();
@@ -397,11 +398,12 @@ void main() {
         originalCompleter.operation.cancel();
         workCompleter.complete(0);
         expect(operation.isCanceled, true);
+        await workCompleter.future;
       });
 
       test('after the value is completed invokes `onValue`', () {
         onValue = expectAsync1((_) => 'foo', count: 1);
-        onCancel = null;
+        onCancel = expectAsync1((_) => '', count: 0);
         originalCompleter.complete(0);
         originalCompleter.operation.cancel();
         var operation = runThen();
--- a/test/cancelable_operation_test.dart
+++ b/test/cancelable_operation_test.dart
@@ -267,10 +267,10 @@ void main() {
       onError = expectAsync2((e, s) => 'Fake', count: 0, id: 'onError');
       onCancel = expectAsync0(() => 'Fake', count: 0, id: 'onCancel');
       propagateCancel = false;
+      originalCompleter = CancelableCompleter();
     });
 
     CancelableOperation<String> runThen() {
-      originalCompleter = CancelableCompleter();
       return originalCompleter.operation.then(onValue!,
           onError: onError,
           onCancel: onCancel,
@@ -386,6 +386,27 @@ void main() {
 
         expect(runThen().value, throwsA('error'));
         originalCompleter.operation.cancel();
+      });
+
+      test('after completing with a future does not invoke `onValue`', () {
+        onValue = expectAsync1((_) => '', count: 0);
+        onCancel = null;
+        var operation = runThen();
+        var workCompleter = Completer<int>();
+        originalCompleter.complete(workCompleter.future);
+        originalCompleter.operation.cancel();
+        workCompleter.complete(0);
+        expect(operation.isCanceled, true);
+      });
+
+      test('after the value is completed invokes `onValue`', () {
+        onValue = expectAsync1((_) => 'foo', count: 1);
+        onCancel = null;
+        originalCompleter.complete(0);
+        originalCompleter.operation.cancel();
+        var operation = runThen();
+        expect(operation.value, completion('foo'));
+        expect(operation.isCanceled, false);
       });
     });
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 2.6.2
+## 2.6.2-dev
 
-* Internal changes only.
+* Fix a bug where `CancelableOperation.then` may invoke the `onValue` callback,
+  even if it had been canceled before `CancelableOperation.value` completes.
 
 ## 2.6.1
 
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -101,7 +101,7 @@ class CancelableOperation<T> {
 
     valueOrCancellation().then((T? result) {
       if (!completer.isCanceled) {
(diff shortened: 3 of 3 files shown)


test/cancelable_operation_test.dart
```diff
@@ -387,6 +387,27 @@ void main() {
...
+      test('after completing with a future does not invoke `onValue`', () {
+        onValue = expectAsync1((_) => '', count: 0);
+        onCancel = null;
+        var operation = runThen();
+        var workCompleter = Completer<int>();
+        originalCompleter.complete(workCompleter.future);
+        originalCompleter.operation.cancel();
+        workCompleter.complete(0);
+        expect(operation.isCanceled, true);
+      });
```
lrhn, line 400: Is it worth adding a delay at the end, perhaps `await workCompleter.future;` to give the work-completer time to actually complete asynchronously, before declaring success.

test/cancelable_operation_test.dart
```diff
@@ -387,6 +387,27 @@ void main() {
         expect(runThen().value, throwsA('error'));
         originalCompleter.operation.cancel();
       });
+
+      test('after completing with a future does not invoke `onValue`', () {
+        onValue = expectAsync1((_) => '', count: 0);
+        onCancel = null;
```
lrhn, line 393: Why not
```dart
 onCancel = expectAsync0(() => 'Fake');
```
It will be called (I assume), so might as well check that it is.

test/cancelable_operation_test.dart
```diff
@@ -387,6 +387,27 @@ void main() {
...
+        var workCompleter = Completer<int>();
+        originalCompleter.complete(workCompleter.future);
+        originalCompleter.operation.cancel();
+        workCompleter.complete(0);
+        expect(operation.isCanceled, true);
+      });
+
+      test('after the value is completed invokes `onValue`', () {
+        onValue = expectAsync1((_) => 'foo', count: 1);
+        onCancel = null;
```
lrhn, line 404: Why not leave it at a `count:0` expectation?

lrhn (approved):

test/cancelable_operation_test.dart
natebosch: We can test either of 2 possibilities. I suppose I could write them as separate test cases but I don't think it adds a lot of safety that we don't already get from other tests.

If `onCancel` is null then I can test that `operation.isCanceled` becomes true.
If `onCancel` is non-null then I can instead test that `operation.isCanceled` stays false, and it completes with the result of `onCancel`.

test/cancelable_operation_test.dart
natebosch: The final `expect` ensures that we have gotten past the point where `onValue` would have been called. I can add this to make the test more clearly correct.

test/cancelable_operation_test.dart
natebosch: Good call, that does add a bit more power to the test. Done.

PR #178, opened 2021-05-08
Make isComplete reflect result availability

### LINKED ISSUE #176: CancelableOperation.isCompleted docs are incorrect
The doc comment for `CancelableOperation.isCompleted` says it is tied to whether the operation can be canceled.

https://github.com/dart-lang/async/blob/5737ad764bfce1ac14918944e7fb538cd674359e/lib/src/cancelable_operation.dart#L134-L139

This isn't true. The operation may still be canceled if the completer was completed with a `Future` that has not yet completed. We want definitely that behavior, otherwise the `CancelableOperation.fromFuture` API makes no sense - you'd never be able to cancel such an operation at all because `isCompleted` is true immediately in that case. I don't think we want to change the behavior of `CancelableCompleter._cancel()`.

https://github.com/dart-lang/async/blob/5737ad764bfce1ac14918944e7fb538cd674359e/lib/src/cancelable_operation.dart#L34-L36

We need to decide if we want `CancelableOparation.isCompleted` to match the first or the second detail in this comment, and update the doc comment, and potentially implementation, to match.

We should either update the doc to reflect current behavior:

```dart
  /// Whether the [CancelableCompleter] backing this operation has been
  /// completed with a value or a Future.
  ///
  /// This value being true does not imply that the [value] future has
  /// completed. If the completer backing this operation was completed
  /// with a Future it may still be possible to cancel this operation.
```

Or, update the doc and change behavior. This would be a breaking change. I think it makes the API more useful, if potentially more confusing.

```dart
  /// Whether the [value] future has completed.
  ///
  /// Listeners on the future may not have been invoked yet,
  /// even if this returns true.
  ///
  /// The [CancelableCompleter] backing this operation may
  /// have been completed, even if this returns false.
  /// That is, `completer.isCompleted` may be true before
  /// before `completer.operation.isCompleted` is true.
  ///
  /// When this returns true, it is no longer possible to
  /// cancel this operation.
```

cc @lrhn

natebosch:
Fixes #176

Change the `CancelableOperation.isComplete` to forward to the `_inner`
completer, which is not completed until the result is available unlike
`CancelableOperation.isComplete` which may lead the result when
`complete` is called with a `Future` argument. Update the docs to
reflect the new behavior. Keep the detail about this property indicating
whether the operation can still be canceled. This detail was incorrectly
stated before, but matches the new implementation.

Remove details pointing to `CancelableCompleter` from the
`CancelableOperation` docs. Flesh out the docs on the completer so that
the distinction between the two `isComplete` getters is more clear.

Remove a detail about not returning `null` from the `cancel()` doc since
the non-nullable return type already makes this clear.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -31,9 +31,7 @@ class CancelableOperation<T> {
   /// It's guaranteed to only be called once.
   ///
   /// Calling this constructor is equivalent to creating a [CancelableCompleter]
-  /// and completing it with [inner]. As such, [isCompleted] is true from the
-  /// moment this [CancelableOperation] is created, regardless of whether
-  /// [inner] has completed yet or not.
+  /// and completing it with [inner].
   factory CancelableOperation.fromFuture(Future<T> inner,
       {FutureOr Function()? onCancel}) {
     var completer = CancelableCompleter<T>(onCancel: onCancel);
@@ -124,20 +122,19 @@ class CancelableOperation<T> {
 
   /// Cancels this operation.
   ///
-  /// This returns the [Future] returned by the [CancelableCompleter]'s
-  /// `onCancel` callback. Unlike [Stream.cancel], it never returns `null`.
+  /// If this operation [isComplete] or [isCanceled] this call is ignored.
+  /// Returns the result of the `onCancel` callback, if one exists.
   Future cancel() => _completer._cancel();
 
   /// Whether this operation has been canceled before it completed.
   bool get isCanceled => _completer.isCanceled;
 
-  /// Whether the [CancelableCompleter] backing this operation has been
-  /// completed.
+  /// Whether the result of this operation is ready.
   ///
-  /// This value being true does not imply that the [value] future has
-  /// completed, but merely that it is no longer possible to [cancel] the
-  /// operation.
-  bool get isCompleted => _completer.isCompleted;
+  /// When ready, the [value] future is completed with the result value
+  /// or error, and this operation can no longer be cancelled.
+  /// An operation may be complete before the listeners on [value] are invoked.
+  bool get isCompleted => _completer._inner.isCompleted;
 }
 
 /// A completer for a [CancelableOperation].
@@ -161,11 +158,18 @@ class CancelableCompleter<T> {
   /// The operation controlled by this completer.
   late final operation = CancelableOperation<T>._(this);
 
-  /// Whether the completer has completed.
+  /// Whether the [complete] or [completeError] have been called.
+  ///
+  /// Once this completer has been completed with either a result or error,
+  /// neither method may be called again.
+  ///
+  /// If [complete] was called with a Future argument, this completer may be
+  /// completed before it's [operation] is completed. In that case the
+  /// [operation] may still be canceled before the result is available.
   bool get isCompleted => _isCompleted;
   bool _isCompleted = false;
 
-  /// Whether the completer was canceled before being completed.
+  /// Whether the completer was canceled before the result was ready.
   bool get isCanceled => _isCanceled;
   bool _isCanceled = false;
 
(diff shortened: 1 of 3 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -174,8 +178,11 @@ class CancelableCompleter<T> {
 
   /// Completes [operation] to [value].
   ///
-  /// If [value] is a [Future], this will complete to the result of that
-  /// [Future] once it completes.
+  /// If [value] is a Future the [operation] will complete with the result of
```
lrhn, line 181: `Future` -> `[Future]`

(Always link source names the first time they occur in a doc, at least code-quote them if they occur more than once).

lib/src/cancelable_operation.dart
```diff
@@ -174,8 +178,11 @@ class CancelableCompleter<T> {
...
+  /// that Future once it is available.
```
lrhn, line 182: `Future` -> `` `Future` ``

lib/src/cancelable_operation.dart
```diff
@@ -202,7 +209,9 @@ class CancelableCompleter<T> {
     });
   }
 
-  /// Completes [operation] to [error].
+  /// Completes [operation] with [error].
```
lrhn, line 212: Doesn't say that it sets `isCompleted` to true, but refers to the value of `isCompleted` instead of whether the completer has been completed.
Prefer to speak in logical terms ("completer has been completed") instead of physical terms ("`isCompleted` is true"). The latter is a consequence of being completed, but then you also have to specify when `isCompleted` becomes true in precise detail.

Instead:
```dart
/// This method may not be called after either [complete] or [completeError] 
/// has been called once.
/// The [isCompleted] is true when either of these methods have been called.
```

Ditto for `completeError`.

lib/src/cancelable_operation.dart
```diff
@@ -212,6 +221,15 @@ class CancelableCompleter<T> {
   }
 
   /// Cancel the completer.
```
lrhn, line 223: ///  Cancels the operation.

Calling `cancel` on the completer ... well, trivially "cancels the completer". That's not saying anything. Logically, it cancels the *operation* that the completer is used to represent.
(And should use active verb.)

lib/src/cancelable_operation.dart
```diff
@@ -212,6 +221,15 @@ class CancelableCompleter<T> {
...
+  ///
+  /// This call will be ignored if the result of the operation is already
```
lrhn, line 225: "will be" -> "is"

Present tense where possible.

lib/src/cancelable_operation.dart
```diff
@@ -212,6 +221,15 @@ class CancelableCompleter<T> {
...
+  /// available.
+  /// This condition may happen some time after [isComplete] is true if
```
lrhn, line 227: Which "condition"? Guessing "result of the operation is .. .available".

Maybe:

/// The result of the operation may only be available some time after
/// the completer has been completed (using [complete] or [completeError],
/// which sets [isCompleted] to true) if completed with a [Future].
/// The completer can be cancelled until the result becomes available,
/// even if [isCompleted] is true.

lib/src/cancelable_operation.dart
```diff
@@ -174,8 +178,11 @@ class CancelableCompleter<T> {
 
   /// Completes [operation] to [value].
   ///
-  /// If [value] is a [Future], this will complete to the result of that
-  /// [Future] once it completes.
+  /// If [value] is a Future the [operation] will complete with the result of
+  /// that Future once it is available.
+  /// In that case [isComplete] will be true before the [operation] is complete.
+  ///
+  /// This method may not be called if [isComplete] is true.
```
lrhn, line 185: Mention that you must only omit the parameter or pass `null` if `T` is nullable.
(That's the requirement inherited from `Completer.complete`. It was a mistake to make that argument optional, but it made sense when everything was nullable).

lib/src/cancelable_operation.dart
lrhn, line 212: `... with [error] and [stackTrakce].`

There is no mention of the `stackTrace` parameter at all.

lrhn (approved):

PR #180, opened 2021-05-20
Add doc comment on Stream.slices

natebosch:
Based on the doc comment on the unmerged Iterable.slices from
https://github.com/dart-lang/collection/pull/191

Drop extra newline in changelog, sort deps, and drop `-nullsafety` from
constraints.

### DIFF
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -6,6 +6,17 @@ import 'dart:async';
 
 /// Utility extensions on [Stream].
 extension StreamExtensions<T> on Stream<T> {
+
+  /// Creates a stream whose elements are contiguous slices of [this].
+  ///
+  /// Each slice is [length] elements long, except for the last one which may be
+  /// shorter if [this] emits too few elements. Each slice begins after the
+  /// last one ends.
+  ///
+  /// For example, `Stream.fromIterable([1, 2, 3, 4, 5]).slices(2)` emits `([1,
+  /// 2], [3, 4], [5])`.
+  ///
+  /// Errors are forwarded to the result stream.
   Stream<List<T>> slices(int length) {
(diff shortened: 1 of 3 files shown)


grouma (approved):

lib/src/stream_extensions.dart
```diff
@@ -6,6 +6,17 @@ import 'dart:async';
...
 /// Utility extensions on [Stream].
 extension StreamExtensions<T> on Stream<T> {
+
+  /// Creates a stream whose elements are contiguous slices of [this].
+  ///
+  /// Each slice is [length] elements long, except for the last one which may be
+  /// shorter if [this] emits too few elements. Each slice begins after the
+  /// last one ends.
+  ///
+  /// For example, `Stream.fromIterable([1, 2, 3, 4, 5]).slices(2)` emits `([1,
```
lrhn, line 16: I don't think you can break code over multiple lines. Break the line before the \`.

lib/src/stream_extensions.dart
```diff
@@ -6,6 +6,17 @@ import 'dart:async';
...
+  /// 2], [3, 4], [5])`.
+  ///
+  /// Errors are forwarded to the result stream.
```
lrhn, line 19: Maybe say:

> Errors are forwarded to the result stream immediately when they occur, 
> even if previous data events have not been emitted because the next slice is not complete yet.

lrhn (approved):

lib/src/stream_extensions.dart
natebosch: Done.

lib/src/stream_extensions.dart
natebosch: Good call on the extra detail. Done.

PR #184, opened 2021-06-13
Fix broadcast StreamGroup ignores streams added after StreamGroup is canceled

### LINKED ISSUE #183: Broadcast StreamGroup doesn't listen to later added streams after _onCancelBroadcast is called
Hi. I'm getting a problem with broadcast `StreamGroup`.

Specifically, when adding a new stream after all previous subscriptions to a broadcast `StreamGroup` have been canceled, the `StreamGroup` won't listen to the newly added stream.

Here's how to reproduce this problem:

```dart

void main() async {
  final streamGroup = StreamGroup<String>.broadcast();

  Stream<String> getFirstStream() async* {
    var i = 0;
    while (true) {
      await Future.delayed(Duration(seconds: 1));
      yield '1st stream: ${i++}';
    }
  }

  // Add the first single subscription stream
  await streamGroup.add(getFirstStream());

  final firstSubs = streamGroup.stream.listen((event) {
    print(event);
  });

  await Future.delayed(Duration(seconds: 3));
  // * Cancel the only subscription to the StreamGroup, putting it into dormant
  await firstSubs.cancel();

  Stream<String> getSecondStream() async* {
    var i = 0;
    while (true) {
      await Future.delayed(Duration(seconds: 1));
      yield '2nd stream: ${i++}';
    }
  }

  // Add another single subscription stream
  await streamGroup.add(getSecondStream());

  // * Start to listen to the StreamGroup again, which should emits events from both the 1st stream and the 2nd stream
  streamGroup.stream.listen((event) {
    print(event);
  });
}
```

### Expected behavior: 
Events from both the 1st stream and the 2nd stream are emitted by the `StreamGroup`.

### Actual behavior: 
Only Events from the 1st stream are emitted. The 2nd stream is ignored.

### Proposal:
I believe the problem is in this line
https://github.com/dart-lang/async/blob/f1c8882c927af68df6786124c94e857de0fa22c1/lib/src/stream_group.dart#L195
Here, since the 1st stream is already being listened and have value `!= null`, the function returns without listening to subsequent streams.

Therefore, it should be `continue` instead of `return`.

Thank you guys for this very helpful package.

anhtuan23:
Fix #183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
   easier to implement custom sinks.
 * Improve performance for `ChunkedStreamReader` by creating fewer internal
   sublists and specializing to create views for `Uint8List` chunks.
+* Don't ignore broadcast streams added to a `StreamGroup` that doesn't have an
+  active listener but previously had listeners and contains a single
+  subscription inner stream.
 
 ## 2.7.0
 
--- a/lib/src/stream_group.dart
+++ b/lib/src/stream_group.dart
@@ -192,7 +192,7 @@ class StreamGroup<T> implements Sink<Stream<T>> {
       // If this is a broadcast group and this isn't the first time it's been
       // listened to, there may still be some subscriptions to
       // single-subscription streams.
-      if (entry.value != null) return;
+      if (entry.value != null) continue;
 
       var stream = entry.key;
       try {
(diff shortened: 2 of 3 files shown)


natebosch:
@googlebot I consent.

kevmoo:
Great find and fix, @anhtuan23 !!

PR #185, opened 2021-06-28
Use BytesBuilder for collectBytes

natebosch:
When this utility was first written `BytesBuilder` was only importable
from `dart:io`. Now that it may be imported from the cross platform
`dart:typed_data` it can be used in `package:async`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 2.7.1-dev
+
 ## 2.7.0
 
 * Add a `Stream.slices()` extension method.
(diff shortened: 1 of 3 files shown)


lrhn (approved):
LGTM!

natebosch:
@lrhn - this broke some clients.

The usage pattern which is broken is creating a `ByteData.view` of the `buffer` from the returned `Uint8List`. I think the correct way to do that is `ByteData.view(bytes.buffer, 0, bytes.lengthInBytes)`, however with the old implementation it would have worked to do `ByteData.view(bytes.buffer)` because the `buffer` has the exact length of the data.

How do you feel about using `toBytes` over `takeBytes` here to allow backwards compatibility for buggy callers?

lrhn:
Could we use a non-copying `BytesBuilder` instead? That would not build the underlying buffer until it needs it.

I'd prefer to fix the buggy code, but if it's widespread, it's probably better to do `toBytes()`.

PR #186, opened 2021-06-30
Avoid copying with BytesBuilder

natebosch:
The previous implementation of `collectBytes` would return a `Uint8List`
where the `lengthInBytes` and `buffer.lengthInBytes` matched. This
property isn't an intended guarantee, however since some callers may
rely on it for correct behavior it is safer to retain it. Use `copy: false`
so that the buffer used in `takeBytes` has the same length as the data.

### DIFF
--- a/lib/src/byte_collector.dart
+++ b/lib/src/byte_collector.dart
@@ -42,11 +42,11 @@ CancelableOperation<Uint8List> collectBytesCancelable(
 /// so it can cancel the operation.
 T _collectBytes<T>(Stream<List<int>> source,
     T Function(StreamSubscription<List<int>>, Future<Uint8List>) result) {
-  var bytes = BytesBuilder();
+  var bytes = BytesBuilder(copy: false);
   var completer = Completer<Uint8List>.sync();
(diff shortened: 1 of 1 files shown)


lib/src/byte_collector.dart
```diff
@@ -46,7 +46,7 @@ T _collectBytes<T>(Stream<List<int>> source,
   var completer = Completer<Uint8List>.sync();
   var subscription =
       source.listen(bytes.add, onError: completer.completeError, onDone: () {
-    completer.complete(bytes.takeBytes());
+    completer.complete(bytes.toBytes());
```
lrhn, line 49: Try using a `BytesBuilder(copy: false)` instead, then use `takeBytes`.

lrhn (approved):
Try the non-copying approach. If that also fails, using `toBytes` LGTM.

lib/src/byte_collector.dart
natebosch: Using `copy: false` seems to fix it. I'll use that. Thanks!

PR #188, opened 2021-07-08
Add *SinkBase classes for implementing custom sinks

nex3:


natebosch (approved):

lib/src/sink_base.dart
```diff
@@ -0,0 +1,168 @@
...
+  /// down. If cleaning up can fail, the error may be reported in the returned
+  /// future.
+  @visibleForOverriding
+  FutureOr<void> onClose();
+
+  /// Asserts that the sink is in a state where adding an event is valid.
+  void _checkCanAddEvent() {
+    if (_closed) throw StateError('Cannot add event after closing');
+  }
+}
```
lrhn, line 57: I think it would be better to have a `SinkBase` which takes `onData`/`onError`/`onDone` as function arguments, or as a single object (could perhaps even be a `StreamSubscription`-like object).

Don't mingle the sink and the response APIs into one interface.

Don't use `@visibleForOverriding`. We generally don't use annotations from `package:meta` in the platform-adjacent packages. (Also not necessary if you don't merge  the interfaces)

lib/src/sink_base.dart
```diff
@@ -0,0 +1,168 @@
...
+/// An abstract class that implements `dart:io`'s [IOSink]'s API in terms of
+/// [onData], [onError], [onClose], and [onFlush] methods.
+///
+/// Because [IOSink] is defined in `dart:io`, this can't officially implement
+/// it. However, it's designed to match its API exactly so that subclasses can
+/// implement [IOSink] without any additional modifications.
+///
+/// This takes care of ensuring that events can't be added after [close] is
+/// called or during a call to [onStream].
+abstract class IOSinkBase extends StreamSinkBase<List<int>> {
```
lrhn, line 107: Not sure this class belongs in `package:async` *because* we can't depend on `dart:io`. I'd create a `package:io` instead if anything.

pubspec.yaml
```diff
@@ -12,6 +12,7 @@ dependencies:
   meta: ^1.1.7
 
 dev_dependencies:
+  charcode: ^1.3.0
```
lrhn, line 15: Don't depend on charcode! it's not a core Dart package!

You can use `package:charcode` to generate the constants you need instead.

lib/src/sink_base.dart
```diff
@@ -0,0 +1,168 @@
...
+    _checkCanAddEvent();
+    onError(error, stackTrace);
+  }
+
+  /// A method that handles error events that are passed to the sink.
+  @visibleForOverriding
+  void onError(Object error, [StackTrace? stackTrace]);
+
+  @override
+  Future<void> close() => _closeMemo.runOnce(onClose);
```
lrhn, line 42: Having an `EventSink` return a `Future` is dangerous since the return type of `EventSink.close` is `void`. That means that *no code expecting an `EventSink` will handle the future*.
Returning a future here is misleading and likely to cause problems for users.

lib/src/sink_base.dart
nex3: I originally tried adding this to `package:io`, but it wants to share so much implementation with `StreamSinkBase` it would have produced a ton of duplication.

Also, I don't think `IOSink` belongs in `dart:io` in the first place, since nothing about its API requires or implies access to system IO. See https://github.com/dart-lang/sdk/issues/17293.

lib/src/sink_base.dart
nex3: Why not use `package:meta` and `@visibleForOverriding` if they result in better APIs? Using inheritance requires the creation of fewer intermediate objects, which results in less complexity and more efficiency.

"Platform-adjacent packages" are only what you declare them to be, and `meta` is certainly a reasonable inclusion since it tightly integrates into the platform.

pubspec.yaml
nex3: Why not? This isn't even a normal dependency, it's only used for tests.

lrhn: Dart-core packages should not depend on non-dart-core packages (unless absolutely necessary and with good arguments).
If we do another breaking change like null safety, we don't want our tests to be stuck on waiting for migration of a package that doesn't have similar support promises to the the core package itself. Even if it's only a dev-dependency, we'd still potentially be held back in migrating the tests, which are just as important as the rest of the package.

lib/src/sink_base.dart
lrhn: I think annotation-based "language extensions" like these are fine for applications, but also that they do not belong in reusable package APIs. Such an API should not expose something that you are not intending to *actually* expose, because users can, and will, use them as if they were just public.
So, it's not a *better* API to me, it's a messier and less structured API, one which mixes different responsibilities in the same interface.

PR #190, opened 2021-07-22
Prepare to publish (again)

natebosch:
I made the previous publish at a commit too early. Move the changelog
for the `StreamGroup` bug fix to a new version and prepare to publish.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,15 @@
+## 2.8.1
+
+* Don't ignore broadcast streams added to a `StreamGroup` that doesn't have an
+  active listener but previously had listeners and contains a single
+  subscription inner stream.
+
 ## 2.8.0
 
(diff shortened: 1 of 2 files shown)


kevmoo (approved):
So you documented a feature in 2.8.0 that WASN'T in that release? Just making sure I understand.

natebosch:
> So you documented a feature in 2.8.0 that WASN'T in that release? Just making sure I understand.

No. `2.8.0` is a valid release with a coherent changelog. We weren't using the `-dev` strategy in this case so publishing from either https://github.com/dart-lang/async/commit/4a1e1d8b912f4121f120205e3cc16a7ddc463e6d or https://github.com/dart-lang/async/commit/25a7e2ec39c03622b86918cb9ce3e7d00dd283d1 would be valid. I published from the earlier commit so neither the bug fix nor the changelog for the bug fix were present.

PR #192, opened 2021-07-26
Adds `EventSinkAdapter`, `StreamSinkAdapter`.

lrhn:
New classes are eplacement for `EventSinkBase`, `StreamSinkBase`,
which are deprecated, and `IOSinkBase` is deprecated with no
replacement. A replacement would belong in a different package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.8.2
+
+* Deprecate `EventSinkBase`, `StreamSinkBase`, `IOSinkBase`.
+* Add `EventSinkAdapter`, `StreamSinkAdapter`.
+
 ## 2.8.1
 
 * Don't ignore broadcast streams added to a `StreamGroup` that doesn't have an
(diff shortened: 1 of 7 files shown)


lrhn:
This is a quick proposal for an API I think is a better fit than the current `EventSinkBase` and `StreamSinkBase`.
WDYT?

natebosch:
I do like this API. Sorry for publishing the prior version before you had a chance to review.

@nex3 - can you confirm that this design works for your use case?

lrhn:
In practice, this CL would probably have to be released as 2.9.0, then the deprecated code can be removed in 2.10.0. So, some version-twiddling neceesary.

lrhn:
I'm not sure the `*Base` classes are general enough that I think it's worth having them in the package (and the extra public-but-not-really members are not what I'd want as an API design).
Could you consider just moving them to the project where you are using them, and not publish them in `dart:async`?

nex3:
I guess we just disagree then. I'm not an owner of this package anymore, so I suppose the ultimate decision is up to you.

lrhn:
I think we'll deprecate the classes for now then, and consider whether we can make a 3.0.0 release without them eventually.

Issue #117, opened 2020-06-09
Include package:stream_transformers into package:async


lrhn:
The package [stream_transformers](/dart-lang/stream_transform/) was created as an experiment. The API has now matured and stabilized, and it might make sense to include it in `package:async`, rather than having two different packages.

It could be exported as a separate top-level library, to avoid the extension methods crowding the classes for users of the rest of `package:async`.

@natebosch

natebosch:
A strong argument in favor of making it a separate top level import is that it will be non-breaking that way. If we were to suddenly introduce these extensions from `package:async/async.dart` then it would introduce errors for everyone who today imports both.

There are no types introduced (publicly) by that package, so copy/paste is a valid migration path.

lrhn:
We would, at the same time, change `package:stream_transform` to export the extensions from `package:async` and therefore depend on a version of `package:async` great enough to have those.

It's still possible to get an older version of `package:stream_transform` and a new version of `package:async`, but it is possible to avoid it.

PR #195, opened 2021-08-27
Add StreamExtensions.firstOrNull

nex3:
By analogy to IterableExtensions.firstOrNull in the collection package.

### DIFF
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -34,8 +34,13 @@ extension StreamExtensions<T> on Stream<T> {
     }));
   }
 
-  /// Returns the first data or error event emitted by [this] if it emits any,
-  /// or `null` if it emits a done event first.
+  /// A future which completes with the first event of this stream, or with
+  /// `null`.
+  ///
+  /// This stream is listened to, and if it emits any event, whether a data
+  /// event or an error event, the future completes with the same data value or
+  /// error. If the stream ends without emitting any events, the future is
+  /// completed with `null`.
   Future<T?> get firstOrNull {
     var completer = Completer<T?>.sync();
     final subscription = listen(null);
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -33,4 +33,21 @@ extension StreamExtensions<T> on Stream<T> {
       sink.close();
     }));
   }
+
+  /// Returns the first data or error event emitted by [this] if it emits any,
+  /// or `null` if it emits a done event first.
+  Future<T?> get firstOrNull {
+    var completer = Completer<T?>.sync();
+    late StreamSubscription<T> subscription;
+    subscription = listen((event) {
+      subscription.cancel();
+      completer.complete(event);
+    }, onError: (Object error, StackTrace stackTrace) {
+      subscription.cancel();
+      completer.completeError(error, stackTrace);
+    }, onDone: () {
+      completer.complete(null);
+    });
+    return completer.future;
+  }
 }
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -38,16 +38,19 @@ extension StreamExtensions<T> on Stream<T> {
   /// or `null` if it emits a done event first.
   Future<T?> get firstOrNull {
     var completer = Completer<T?>.sync();
-    late StreamSubscription<T> subscription;
-    subscription = listen((event) {
-      subscription.cancel();
-      completer.complete(event);
-    }, onError: (Object error, StackTrace stackTrace) {
-      subscription.cancel();
-      completer.completeError(error, stackTrace);
-    }, onDone: () {
-      completer.complete(null);
-    });
+    final subscription = listen(null);
+    subscription
+      ..onData((event) {
+        subscription.cancel();
+        completer.complete(event);
+      })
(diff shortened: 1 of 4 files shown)


natebosch:
I think we need to hold off on this until we have resolved https://github.com/dart-lang/async/issues/117, and then we can put this in that library.

In the short term we could add this in `package:stream_transform` which will have a (hopefully) smooth transition once we do incorporate it here.

lib/src/stream_extensions.dart
```diff
@@ -33,4 +33,21 @@ extension StreamExtensions<T> on Stream<T> {
...
+    late StreamSubscription<T> subscription;
+    subscription = listen((event) {
+      subscription.cancel();
+      completer.complete(event);
+    }, onError: (Object error, StackTrace stackTrace) {
+      subscription.cancel();
+      completer.completeError(error, stackTrace);
+    }, onDone: () {
+      completer.complete(null);
+    });
```
natebosch, line 50: For wherever we end up landing this...

I think we can avoid the `late`

```suggestion
    final subscription = listen(null);
    subscription
      ..onData((event) {
        subscription.cancel();
        completer.complete(event);
      })
      ..onError((Object error, StackTrace stackTrace) {
        subscription.cancel();
        completer.completeError(error, stackTrace);
      })
      ..onDone(() {
        completer.complete(null);
      });
```

nex3:
I don't think this would make sense in `stream_transformer`, since it's not representable as a transformer—it's a stream-to-future transformation, not a stream-to-stream transformation. Also, there's precedent for adding new features here rather than there in #175.

natebosch:
> it's not representable as a transformer—it's a stream-to-future transformation, not a stream-to-stream transformation

That's a good point.

@lrhn - which library would you prefer this goes into?

lib/src/stream_extensions.dart
```diff
@@ -33,4 +33,24 @@ extension StreamExtensions<T> on Stream<T> {
       sink.close();
     }));
   }
+
+  /// Returns the first data or error event emitted by [this] if it emits any,
```
lrhn, line 37: Should be documented with a noun phrase (it's a getter).

So either (underplaying the asynchrony)

/// The first data or event emitted by this stream, or `null` if the stream is empty.

or mention the future,

/// A future which completes with the first event of this stream, or with `null`.
///
/// If this stream emits any event, whether a data event or an error event,
/// the future completes with the same data value or error.
/// If the stream ends without emitting any events, the future is completed with
/// the `null` value.

I think I prefer the latter.

lib/src/stream_extensions.dart
```diff
@@ -33,4 +33,24 @@ extension StreamExtensions<T> on Stream<T> {
...
+  /// or `null` if it emits a done event first.
+  Future<T?> get firstOrNull {
+    var completer = Completer<T?>.sync();
+    final subscription = listen(null);
```
lrhn, line 41: Can use `cancelOnError: true`. Then you can just tear off `completer.completeError`instead of creating a new closure. It can even be:

```dart
  var subscription = listen(null, 
    onError: completer.completeError,
    onDone: () { 
      completer.complete(null);
    },
    cancelOnError: true);
  subscription.onData((event) {
    subscription.cancel();
    completer.complete(event);
  });
```

test/stream_extensions_test.dart
```diff
@@ -55,4 +55,30 @@ void main() {
...
+
+    test('returns the first error event', () {
+      expect(Stream.error('oh no').firstOrNull, throwsA('oh no'));
+    });
+
+    test('returns null for an empty stream', () {
+      expect(Stream.empty().firstOrNull, completion(isNull));
+    });
+
+    test('cancels the subscription after an event', () async {
```
lrhn, line 73: Could check that it also happens after an error. It is a different code path.

lrhn (approved):
I see no problem with adding it to `StreamExtensions` in `package:async`. It's the perfect place. The corresponding collection method was in `package:collection`, so the analogy is perfect.

Code looks fine too.

We could consider adding more extensions along with this, perhaps before releasing 2.9.0, but starting here seems fine.

lib/src/stream_extensions.dart
nex3: I considered doing it that way, but I think the symmetry between the `onData` and `onError` events is easier to follow than having one of them manually cancel the subscription and the other cancel it through `cancelOnError`.

natebosch (approved):

lib/src/stream_extensions.dart
```diff
@@ -33,4 +33,29 @@ extension StreamExtensions<T> on Stream<T> {
...
+  /// This stream is listened to, and if it emits any event, whether a data
+  /// event or an error event, the future completes with the same data value or
+  /// error. If the stream ends without emitting any events, the future is
+  /// completed with `null`.
+  Future<T?> get firstOrNull {
+    var completer = Completer<T?>.sync();
+    final subscription = listen(null);
+    subscription
+      ..onData((event) {
+        subscription.cancel();
```
lrhn, line 49: I did forget to say that we should wait for the cancel to complete before competing the future. :frowning_face: 

Using `cancelOnError` ensures that the cancel completes before the `onError` callback.
We'll have to do that manually for the `onData` cancel.

natebosch: #196

PR #198, opened 2021-10-01
update some StreamGroup apis to Future<void> from Future<dynamic>

jakemac53:
This is probably technically breaking for some cases - but it doesn't look to be breaking for anything at all internally, so I think in practice it is not breaking. Finishing up a global presubmit now with all the changes (but already did one with a subset).

WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
-## 2.9.0
+## 2.9.0-dev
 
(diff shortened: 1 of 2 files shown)


natebosch (approved):

lrhn (approved):
I think it's a fine change. It's unlikely to break anything because no-one should be looking at the value anyway - because there is no value.

PR #206, opened 2022-03-09
Avoid leaking unreachable listeners on `CancelableOperation`.

### LINKED ISSUE #200: Possible memory leaks?
If the operation was canceled before it completes then [operation.value](https://github.com/dart-lang/async/blob/2.8.1/lib/src/cancelable_operation.dart#L43) will never complete so the resgistered callbacks / error callbacks will just hang in memory and if they reference some huge objects then they also won't be garbage collected.

```dart
  var opration = CancelableOperation.fromFuture(
    Future(() => 1),
    onCancel: () => print('onCancel'),
  );

  Future.sync(() => opration.cancel());

  opration.value.then((value) {
    // Holding some reference to Huge object
  }, onError: (e, s) {
    // Holding some reference to Huge object
  });

  opration.value.whenComplete(() {
    // Holding some reference to Huge object
  });

  opration.asStream().listen(
    (event) {
      // Holding some reference to Huge object
    },
    onDone: () => {
      // Holding some reference to Huge object
    },
  );
```

You might suggest I should use [valueOrCancellation](https://github.com/dart-lang/async/blob/2.8.1/lib/src/cancelable_operation.dart#L69) however it also registers some [callback functions](https://github.com/dart-lang/async/blob/2.8.1/lib/src/cancelable_operation.dart#L71-L72) which will stay alive since [value won't be completed](https://github.com/dart-lang/async/blob/2.8.1/lib/src/cancelable_operation.dart#L207-L213).

`asStream` also registers some [callback functions](https://github.com/dart-lang/async/blob/2.8.1/lib/src/cancelable_operation.dart#L53-L59).

lrhn:
Make `CancelableOperation` not hold onto unnecessary callbacks.
    
The existing `CancelableOperation` contains two completers,
where at most one will ever complete. Listeners added to
the other completer's future will never be released as
long as the operation is alive, even after the operation
has otherwise completed.
    
This change drops reference to the other completer when
one completer is chosen for completion.
That can allow the completer, its future, and that future's listeners
to be GC'ed when it's known that they'll never be relevant again.
    
The change has one visible effect: If asking for the `value`
future after the operation has been cancelled and the value
completer has been released, you'll get a new future,
which also never completes, which is not identical to
other futures returned by the same getter.
This only matters if someone checks the future for identity,
their behavior is exactly the same (no behavior, whatsoever).
    
Since the same class returns a completely new future on each
call read of `valueOrCancellation`, I think that's a reasonable
change, and unlikely to affect anyone.
(Users are not usually expecting asynchronous functions to
return the same future every time since `async` functions
don't.)
    
Fixes #200.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -237,25 +237,21 @@ class CancelableCompleter<T> {
   /// The callback to call if the operation is canceled.
   final FutureOr<void> Function()? _onCancel;
 
-  /// Whether `complete` or `completeError` may still be called.
+  /// Whether [complete] or [completeError] may still be called.
   ///
   /// Set to false when calling either.
+  ///
+  /// When completing by calling [complete] with a future,
+  /// it's still possible to cancel until the result is actually
+  /// available.
+  /// You are also allowed to call [complete] or [completeError]
+  /// after the operation has been canceled, as long as you only call it once.
+  /// It just won't do anything after the operation is cancelled.
+  /// This value only guards the calls to [complete] and [completeError].
   bool _mayComplete = true;
 
   /// The operation controlled by this completer.
   late final operation = CancelableOperation<T>._(this);
-
-  /// Becomes true when [complete] or [completeError] is called.
-  ///
-  /// Completing twice is not allowed.
-  ///
-  /// If [complete] is called with a future, it's still possible to
-  /// cancel the operation until that future completes,
-  /// so this value and [_isCanceled] are not mutually exclusive.
-  bool get _isCompleted => _cancelCompleter == null;
-
-  /// Whether the completer was canceled before the result was ready.
-  bool get _isCanceled => _inner == null;
 
   /// Creates a new completer for a [CancelableOperation].
   ///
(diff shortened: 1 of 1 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -103,13 +103,16 @@ class CancelableOperation<T> {
...
 
-    value.then((value) {
-      controller.add(value);
-      controller.close();
-    }, onError: (Object error, StackTrace stackTrace) {
-      controller.addError(error, stackTrace);
-      controller.close();
-    });
+    var inner = _completer._inner;
+    if (inner != null) {
```
natebosch, line 107: [optional] Could reduce indentation with `inner?.future.then` or `_completer._inner?.future.then`

natebosch (approved):
Should we add a changelog?

lib/src/cancelable_operation.dart
lrhn: Good point!

lrhn:
I don't think a changelog entry is required. There is no user-visible change, it's just an optimization (a memory optimization in this case).

PR #207, opened 2022-03-14
Make ephemeral async cache clear sooner.

### LINKED ISSUE #151: Ephemeral AsyncCache is not truly ephemeral
Hi,

Here is a repro:

```dart

Future<void> main(List<String> arguments) async {
  final cache = AsyncCache.ephemeral();

  final firstValue = await cache.fetch(_getValue);
  print('First value: $firstValue');
  final secondValue = await cache.fetch(_getValue);
  print('Second value: $secondValue');
}

int _value = 0;
Future<int> _getValue() async {
  await Future<void>.delayed(const Duration(milliseconds: 500));
  return ++_value;
}
```

This outputs:

```
First value: 1
Second value: 1
```

My expectation would be that the cached future would be invalidated immediately after the first `fetch` is awaited, leading to this output:

```
First value: 1
Second value: 2
```

The reason this doesn't happen is because an ephemeral `AsyncCache` has a `duration` of zero, but invalidation always uses a `Timer`, so invalidation won't occur until the event loop gets a chance to execute (after my `main`).

By making the following adjustments to `AsyncCache`, I get the desired behavior:

```dart
  Future<T> fetch(Future<T> Function() callback) async {
    if (_cachedStreamSplitter != null) {
      throw StateError('Previously used to cache via `fetchStream`');
    }

    var future = _cachedValueFuture;

    if (future == null) {
      future = _cachedValueFuture = callback();
      await future;
      _startStaleTimer();
    }

    return future;
  }

  void _startStaleTimer() {
    if (_duration == Duration.zero) {
      invalidate();
    } else {
      _stale = Timer(_duration, invalidate);
    }
  }
```
I believe `fetchStream` would require a tweak similar to that I made to `fetch`.

Am I right in thinking this is a bug, or is my expectation of behavior incorrect?

lrhn:
The `AsyncCache.ephemeral()` cache would invalidate
the `fetch` cache after a `Duration.zero` wait.
That allows a large number of microtasks to happen between
the `fetch` completing and the cache being invalidated.

Instead the cache is now invalidated immediately when the fetched
request completes.

Fixes #151

### DIFF
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -26,7 +26,10 @@ import 'package:async/async.dart';
 /// [fake_async]: https://pub.dev/packages/fake_async
 class AsyncCache<T> {
   /// How long cached values stay fresh.
-  final Duration _duration;
+  ///
+  /// Set to `null` for ephemeral caches, which only stay alive until the
+  /// future completes.
+  final Duration? _duration;
 
   /// Cached results of a previous [fetchStream] call.
   StreamSplitter<T>? _cachedStreamSplitter;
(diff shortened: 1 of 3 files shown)


lib/src/async_cache.dart
```diff
@@ -60,12 +63,8 @@ class AsyncCache<T> {
...
       throw StateError('Previously used to cache via `fetchStream`');
     }
-    final result = _cachedValueFuture ??= callback();
-    try {
-      return await result;
-    } finally {
-      _startStaleTimer();
-    }
+    return _cachedValueFuture ??= callback()
+      ..whenComplete(_startStaleTimer).ignore();
```
natebosch, line 67: Is this behavior change intentional? Do we still want to call `_startStaleTimer` if `_cachedValueFuture` was non-null?
```suggestion
    return (_cachedValueFuture ??= callback())
      ..whenComplete(_startStaleTimer).ignore();
```

natebosch (approved):

lib/src/async_cache.dart
lrhn: I think this is actually correct, and the original wasn't. The original could trigger multiple invalidations (at the same time, `_duration` after the future completed, so probably not a problem). One is sufficient.

lrhn:
I believe the earlier change was fixed, so landing.

Issue #210, opened 2022-03-17
It is not easy to chain a CancelableOperation following other work


natebosch:
There is no equivalent of `async` and `await`, and the chaining we have through `.then` requires `FutureOr` returns from the callbacks - we can't do any cancelable work inside the callback. These factors mean that some idioms which feel easy with `Future` feel clunky, and require careful use of `CancelableCompleter`.

For example, I tried to write a retry when a `CancelableOperation` results in an error. The nicest I could come up with is:

```dart
CancelableOperation<T> retryOperation<T>(
    CancelableOperation<T> Function() callback,
    {int allowedTries = 3}) {
  late CancelableOperation<T> operation;
  final completer = CancelableCompleter<T>(onCancel: () => operation.cancel());
  _retry(() => operation = callback(), completer, allowedTries);
  return completer.operation;
}

Future<void> _retry<T>(CancelableOperation<T> Function() callback,
    CancelableCompleter<T> completer, int remainingTries) async {
  for (;;) {
    try {
      final current = callback();
      var result = await current.valueOrCancellation();
      if (completer.isCanceled) return;
      if (current.isCompleted && !current.isCanceled) {
        completer.complete(result);
      } else {
        completer.operation.cancel();
      }
      return;
    } on Object catch (error, stackTrace) {
      if (--remainingTries > 0) continue;
      if (completer.isCanceled) return;
      completer.completeError(error, stackTrace);
    }
  }
}
```

In another place I tried to wrap a `CancelableOperation Function()` where it needs some async work before forwarding to a delegate, which is equivalent to flattening a `Future<CancelableOperation>` to a `CancelableOperation`.

```dart
CancelableOperation<T> flatten<T>(
    Future<CancelableOperation<T>> eventualOperation) {
  late final CancelableOperation<T> operation;
  final completer = CancelableCompleter<T>(onCancel: () => operation.cancel());
  eventualOperation.then((result) {
    operation = result;
    return result.valueOrCancellation().then((innerResult) {
      if (completer.isCanceled) return;
      if (operation.isCompleted && !operation.isCanceled) {
        completer.complete(innerResult);
      } else {
        completer.operation.cancel();
      }
    });
  }, onError: (Object error, StackTrace stackTrace) {
    if (completer.isCompleted) return;
    completer.completeError(error, stackTrace);
  });
  return completer.operation;
}
```

[...]

PR #211, opened 2022-03-18
Add CancelableOperation.thenOperation

natebosch:
Towards #210

Once combined with `Completer.completeOperation` this signature adds
significant flexibility for chaining Cancelable work following other
async work.

Move the existing `.then` implementation to `.thenOperation` since the latter
is more general.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -264,7 +264,8 @@ class CancelableOperation<T> {
                 try {
                   await onError(error, stack, completer);
                 } catch (error2, stack2) {
-                  completer.completeError(error2, stack2);
+                  completer.completeError(
+                      error2, identical(error, error2) ? stack : stack2);
                 }
               });
     _completer._cancelCompleter?.future.whenComplete(onCancel == null
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -150,6 +150,65 @@ class CancelableOperation<T> {
       {FutureOr<R> Function(Object, StackTrace)? onError,
       FutureOr<R> Function()? onCancel,
       bool propagateCancel = false}) {
+    if (onError != null && onCancel != null) {
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, onError: (error, stack, completer) {
+        completer.complete(onError(error, stack));
+      }, onCancel: (completer) {
+        completer.complete(onCancel());
+      }, propagateCancel: propagateCancel);
+    } else if (onError != null) {
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, onError: (error, stack, completer) {
+        completer.complete(onError(error, stack));
+      }, propagateCancel: propagateCancel);
+    } else if (onCancel != null) {
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, onCancel: (completer) {
+        completer.complete(onCancel());
+      }, propagateCancel: propagateCancel);
+    } else {
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, propagateCancel: propagateCancel);
+    }
+  }
+
+  /// Creates a new cancelable operation to be completed
+  /// when this operation completes or is cancelled.
+  ///
+  /// The [onValue] and [onError] callbacks behave in the same way as
+  /// for [Future.then], and the result of those callbacks is used to complete
+  /// the returned cancelable operation.
+  ///
+  /// If this operation completes normally [onValue] is called with the result
+  /// and the [CancelableCompleter] controlling the returned operation.
+  ///
+  /// If this operation completes as an error, and an [onError] callback is
+  /// provided, it is called with the error, stack trace, and the
+  /// [CancelableCompleter] controlling the returned operation.
+  ///
+  /// If this operation is canceled, and an [onCancel] call is provided, it is
+  /// called with the [CancelableCompleter] controlling the returned operation.
+  ///
+  /// If [onCancel] is not provided, and this operation is canceled, then the
+  /// returned operation is also canceled.
+  ///
+  /// At most one of [onValue], [onError], or [onCancel] will be called.
+  /// If any of [onValue], [onError], or [onCancel] cause any asynchronous
+  /// errors, including through a returned `Future`, they will not be handled or
+  /// forwarded through the returned operation.
+  ///
+  /// If [propagateCancel] is `true` and the returned operation is canceled
+  /// then this operation is canceled. The default is `false`.
+  CancelableOperation<R> thenOperation<R>(
+      void Function(T, CancelableCompleter<R>) onValue,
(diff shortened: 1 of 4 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -150,6 +150,65 @@ class CancelableOperation<T> {
...
+  /// If [onCancel] is not provided, and this operation is canceled, then the
+  /// returned operation is also canceled.
+  ///
+  /// At most one of [onValue], [onError], or [onCancel] will be called.
+  /// If any of [onValue], [onError], or [onCancel] cause any asynchronous
+  /// errors, including through a returned `Future`, they will not be handled or
+  /// forwarded through the returned operation.
+  ///
+  /// If [propagateCancel] is `true` and the returned operation is canceled
+  /// then this operation is canceled. The default is `false`.
```
natebosch, line 206: @lrhn  I tried to go back and understand the motivation for this based on https://github.com/dart-lang/async/pull/81#issuecomment-479444606 but I'm not sure I understand.

I can't think of the reason you _wouldn't_ want to propagate cancellation back to `this` from the returned operation. I was tempted to not have this argument, or change the default, because in each usage I've encountered so far it's felt like extra noise. Would it be better to keep consistency even if most use cases need to pass the argument?

Do you recall what use case we expected where you wouldn't want the cancellation to propagate?

lib/src/cancelable_operation.dart
```diff
@@ -150,6 +150,65 @@ class CancelableOperation<T> {
...
+  /// If this operation is canceled, and an [onCancel] call is provided, it is
+  /// called with the [CancelableCompleter] controlling the returned operation.
+  ///
+  /// If [onCancel] is not provided, and this operation is canceled, then the
+  /// returned operation is also canceled.
+  ///
+  /// At most one of [onValue], [onError], or [onCancel] will be called.
+  /// If any of [onValue], [onError], or [onCancel] cause any asynchronous
+  /// errors, including through a returned `Future`, they will not be handled or
+  /// forwarded through the returned operation.
```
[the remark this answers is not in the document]
natebosch: I gave this a try. Let me know what you think.

lib/src/cancelable_operation.dart
lrhn: No recollection. My guess is that it was added later, and therefore the default was made the previous behavior.

We can do a code search to see if anyone ever uses the default. If everybody passes `true` anyway, we might get away with changing the default (in a major version increment, obviously).

natebosch: >  My guess is that it was added later, and therefore the default was made the previous behavior.

That was my original assumption, but it turns out the parameter and default were there from the start. The only discussion I can find about propagation is the thread I linked above (Also mentioned [here](https://github.com/dart-lang/async/pull/83#issuecomment-480074557) but without any new context). It has a bit of discussion of both forwards propagation (`this` gets cancelled so  the returned operation gets canceled) and backwards propagation (the `propagateCancel` behavior), but I don't understand how exactly the discussion lead to the code we shipped.

> We can do a code search to see if anyone ever uses the default.

There are zero usages I can find of the argument being passed, either internally or on github. (Maybe no one really cares?)

There are many uses where I don't think it matters.
Two common patterns are:

- `return CancelableOperation.fromFuture(someFuture).then(transformResult)` - the cancelation propagation doesn't have an observable impact. Only the transform work can be canceled, `someFuture` happens either way.
- `someOperation().then(useResult);` - the returned operation is dropped and can't be cancelled, so propagation never matters. Could have been written as `someOperation().value.then(useResult)` but the former is shorter.

I also see uses where I suspect the cancellation propagation is what was intended, but was maybe never tested. `return someOperation().then(transformResult);`.  Since `someOperation()` returns a `CancelableOperation` I'd expect the intention is for the caller to be able to cancel it - and this caller will never cancel it (but the author might think they are).

I have not yet seen a case that strikes me as one where an operation could have multiple uses that need to avoid canceling each other.

I can try changing the default and see if there happen to be any tests for relying on this behavior.

lib/src/cancelable_operation.dart
```diff
@@ -150,6 +150,65 @@ class CancelableOperation<T> {
...
+      }, propagateCancel: propagateCancel);
+    } else {
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, propagateCancel: propagateCancel);
+    }
+  }
+
+  /// Creates a new cancelable operation to be completed
+  /// when this operation completes or is cancelled.
```
lrhn, line 181: This is a long first paragraph (guildeline is one line).
Not that I can find a good shorter version, though.

lib/src/cancelable_operation.dart
```diff
@@ -182,7 +241,7 @@ class CancelableOperation<T> {
             : (Object error, StackTrace stack) {
                 if (completer.isCanceled) return;
                 try {
-                  completer.complete(onError(error, stack));
+                  onError(error, stack, completer);
                 } catch (error2, stack2) {
                   completer.completeError(error2, stack2);
```
lrhn, line 258: So, here, if we want to work exactly like `Future.then`, we can do:
```dart
 completer.completeError(error2, identical(error, error2) ? stack : stack2);
```
(It's a silly thing, and these days you can just do `Error.throwWithStackTrace(e, s)` to explicitly ask for the same stack trace, so the functionality isn't strictly necessary.)

lib/src/cancelable_operation.dart
```diff
@@ -150,6 +150,65 @@ class CancelableOperation<T> {
...
+        completer.complete(onValue(value));
+      }, propagateCancel: propagateCancel);
+    }
+  }
+
+  /// Creates a new cancelable operation to be completed
+  /// when this operation completes or is cancelled.
+  ///
+  /// The [onValue] and [onError] callbacks behave in the same way as
+  /// for [Future.then], and the result of those callbacks is used to complete
```
lrhn, line 184: same way ... well, almost. `Future.then` is a little clever. If the `onError` handler throws the *same* error object that it got in, it considers it a rethrow and uses the old stack instead of the new.
I don't actually think that's valuable here, but maybe change "in the same way as" to "similarly to".

lib/src/cancelable_operation.dart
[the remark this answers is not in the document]
lrhn: LGTM. That does mean that throwing asynchronously becomes an uncaught error.
So, just don't do that then :)

lrhn (approved):

lib/src/cancelable_operation.dart
```diff
@@ -150,6 +150,65 @@ class CancelableOperation<T> {
...
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, onCancel: (completer) {
+        completer.complete(onCancel());
+      }, propagateCancel: propagateCancel);
+    } else {
+      return thenOperation<R>((value, completer) {
+        completer.complete(onValue(value));
+      }, propagateCancel: propagateCancel);
+    }
```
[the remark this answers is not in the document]
natebosch: Good call. Done.

lib/src/cancelable_operation.dart
natebosch: Done.

PR #213, opened 2022-03-23
Change the default propagateCancel argument

### LINKED ISSUE #212: Change the default of `propagateCancel` argument in CancelableOperation.then
See discussion in https://github.com/dart-lang/async/pull/211#discussion_r829662531

I haven't been able to find any tests that fail when I change the default, but it's hard to say for sure if there could be untested behavior.

I'm fairly confident that some of the usage I saw which does not pass the argument would expect cancellation to propagate.

cc @TastyPi - do you recall why you chose not propagating as the default? Do you have any use cases where propagating the cancel would result in incorrect behavior?

cc @lrhn @devoncarew - I'm tempted to try to roll out a change in the default as a minor release. I'm concerned that the unsafe default is going to cause more pain than potentially breaking an existing usage. WDYT?

natebosch:
Closes #212

Authors are more likely to expect propagation, and not cancelling is
often a hidden inefficiency so it often goes unnoticed and untested.

Update the doc comment to focus on the 3 ways an operation can end - as
a value, an error, or a cancellation. This will hopefully make it more
clear that the `onCancel` callback does not relate to the cancellation
of the returned operation.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -133,25 +133,25 @@ class CancelableOperation<T> {
   /// Creates a new cancelable operation to be completed when this operation
   /// completes normally or as an error, or is cancelled.
   ///
-  /// If this operation completes normally the [value] is passed to [onValue]
+  /// If `this` operation completes normally the [value] is passed to [onValue]
   /// and the returned operation is completed with the result.
   ///
-  /// If this operation completes as an error, and no [onError] callback is
+  /// If `this` operation completes as an error, and no [onError] callback is
   /// provided, the returned operation is completed with the same error and
   /// stack trace.
-  /// If this operation completes as an error, and an [onError] callback is
+  /// If `this` operation completes as an error, and an [onError] callback is
   /// provided, the returned operation is completed with the result.
   ///
-  /// If this operation is canceled, and no [onCancel] callback is provided, the
-  /// returned operation is canceled.
-  /// If this operation is canceled, and an [onCancel] callback is provided, the
-  /// returned operation is completed with the result.
+  /// If `this` operation is canceled, and no [onCancel] callback is provided,
+  /// the returned operation is canceled.
+  /// If `this` operation is canceled, and an [onCancel] callback is provided,
+  /// the returned operation is completed with the result.
   ///
-  /// If the returned operation is canceled before this operation completes or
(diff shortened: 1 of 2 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -130,26 +130,33 @@ class CancelableOperation<T> {
...
-  ///
-  /// If [onCancel] is not provided, and this operation is canceled, then the
-  /// returned operation is also canceled.
-  ///
-  /// If [propagateCancel] is `true` and the returned operation is canceled
-  /// then this operation is canceled. The default is `false`.
+  /// Creates a new cancelable operation to be completed when this operation
+  /// completes normally or as an error, or is cancelled.
+  ///
+  /// If this operation completes normally the [value] is passed to [onValue]
```
natebosch, line 136: @lrhn - WDYT about using the term `source operation` or `original operation`? I think `returned operation` is fairly clear. I intended for the consistent usage of `this operation` to convey the "this" that the method is called on, but I'm not sure if that comes across.

natebosch: @jakemac53 - you and I had discussed this briefly and I think you were ok with `this`. Should we consider making it `[this]` or `` `this` ``?

jakemac53: I don't feel strongly, not sure what `[this]` would link to and I would only do that if dartdoc understand it and does something sensible, but `` `this` `` sgtm.

jakemac53 (approved):

lib/src/cancelable_operation.dart
natebosch: It looks like Dartdoc will link `[this]` to the class, which is sensible but maybe not useful in this situation.

Analyzer does not have a definition to jump to, though I'd imagine if we consider it important we could make it align with dartdoc.

jakemac53: I think just backticks is good then, linking the class isn't very helpful.

lrhn: I'd say no quoting at all. If you use a quoted \`this\`, it refers to an object and should be a noun.
In "this operation", the "this" is not a noun. (I don't know the English word for what it is, though. Preposition?)

And using "this operation" is fine with me,\

lrhn (approved):

PR #215, opened 2022-04-18
Add CancelableCompleter.completeOperation

natebosch:
Towards #210

Combined with `CancelableOperation.thenOperation` this allows chaining
cancelable work that can be canceled at multiple points.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -418,6 +418,33 @@ class CancelableCompleter<T> {
     });
   }
 
+  /// Makes this [CancelableCompleter.operation] complete with the same result
+  /// as [result].
+  ///
+  /// If the operation of this completer is cancelled before [result] completes,
+  /// then if [propagateCancel] is set to true, [result] is also cancelled.
+  void completeOperation(CancelableOperation<T> result,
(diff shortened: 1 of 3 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -418,6 +418,33 @@ class CancelableCompleter<T> {
     });
   }
 
+  /// Makes this [CancelableCompleter.operation] complete with the same result
+  /// as [result].
+  ///
+  /// If the operation of this completer is cancelled before [result] completes,
+  /// then if [propagateCancel] is set to true, [result] is also cancelled.
```
lrhn, line 425: I use "\`true\`" for boolean literals.

The "if... then if... then" can be "if ... and ..., then"
I'd probably put the `propagateCancel` clause first:

> /// If [propagateCancel] is `true` (the default), 
> /// and the [operation] of this completer is cancelled
> /// before [result] completes,
> /// then [result] is also cancelled.

lrhn (approved):

lib/src/cancelable_operation.dart
natebosch: I like that phrasing. Done.

Issue #31, opened 2017-05-15
Add an "Observable" class - wrapper + updates stream


emesx:
Add an `Observable` class (name suggestions welcome!) that would combine a (wrapped) value and a `Stream` of updates made to that value.

@matanlurey was also interested in this.

matanlurey:
/cc some other interested parties: @natebosch @jakemac53.

Yes, I'd like to see this get in. From a high-level, I _think_ we should support:

* An interface that is both a `Stream<T>` + a value of `T`.
* A way to convert a `Stream<T>` into this class where `value` is the previous value.
* A way to change `value` and have that automatically publish on the stream.

_Rough_ API, assuming we call this `Property` (please suggest better names):

```dart
import 'dart:async';

/// Combines a [value] and a stream of updates made to that value.
///
/// This pattern makes it easy to both be using in a dirty-check fashion by exposing
/// a [value] field, and also reactive programming by listening to the property updates:
/// ```
/// void doThing(Property<String> name) {
///   name.listen((v) => print('Your name now is: $v.'));
/// }
/// ```
abstract class Property<T> implements Stream<T> {
  const factory Property.fromStream(Stream<T> stream, [T initialValue]) = _StreamProperty;

  T get value;
}

/// Exposes a [value] setter on top of the interface of [Property] for mutations.
///
/// ```
/// var name = new MutableProperty<String>('Jill User');
/// ...
/// name.value = 'Joe User';
/// ```
abstract class MutableProperty<T> implements Property<T> {
  factory MutableProperty([T initialValue]) = _MutableProperty;

  set value(T value);
}

class _PropertyStream<T> extends StreamView<T> { ... }
class _MutableProperty<T> extends Stream<T> implements MutableProperty<T> { ... }
````

A couple of questions:

- [ ] When providing an `initialValue` (if accepted), should it be emitted on the stream?
- [ ] Should the stream be `broadcast` or single-subscription, or customizable?
- [ ] Should only distinct values be emitted? If so should equality be customizable or just `==`?

matanlurey:
(Offline we discussed `Observable` is too overloaded to be useful here)

emesx:
`Property` is a nice name, I like it (although it doesn't really capture the sense of this class I think).

Regarding the questions, I doubt we'll find a one-size-fits-all solution. Hence, two options: we either implement all the options and expose them via named ctors / factory, or we implement one, most useful one and iterate from there.

One more checkbox to consider:
[...]

PR #218, opened 2022-06-03
Adds `ObservableValue` class for observing state changes.

lrhn:
@macjohnny

### DIFF
--- a/lib/src/observable_value.dart
+++ b/lib/src/observable_value.dart
@@ -99,7 +99,7 @@ class _ObservableValue<T> implements ObservableValue<T> {
     // to only contain listeners which are still active.
 
     // Since `add` is asynchronous, this loop will not call any user code.
-    int activeCount = 0;
+    var activeCount = 0;
     for (var i = 0; i < _listeners.length; i++) {
       var controller = _listeners[i];
       if (controller.hasListener) {
--- a/lib/src/observable_value.dart
+++ b/lib/src/observable_value.dart
@@ -1,0 +1,115 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// A single variable with a stream emitting the variable's changing values.
+///
+/// Contains a single [value] which can be read and written as any other
+/// variable.
+///
+/// Whenever the value is set, the new value is emitted on the
+/// [updates] stream.
+/// Further, any new listener on the [updates] stream is initially sent the
+/// current [value].
+///
+/// Consider using immutable types for [value] since listeners will not
+/// be notified if the internal state of [value] changes, only if
+/// the [value] variable is written to.
+///
+/// Example:
+/// ```dart
+/// class ToggleWidget {
+///   final ObservableValue<bool> _isVisible = ObservableValue(true);
+///
+///   // ...
+///
+///   Stream<bool> get visibilityChanges => _isVisible.updates;
+///
+///   void toggleVisibility() {
+///     _isVisible.value ^= true; // XOR with true toggles value.
+///   }
+/// }
+/// ```
+abstract class ObservableValue<T> {
+  /// The current value.
+  ///
+  /// The [updates] stream emits the current value when listened to,
+  /// and the new value when `value` is set to a new value.
+  abstract T value;
+
+  /// A stream emitting the current value of [value].
+  ///
+  /// New listeners will receive an event containing the current value
+  /// of [value] as soon as possible after listening.
+  /// Every time [value] changes, they will then receive an event
+  /// with the new value.
+  ///
+  /// This stream can be listened to multiple times.
+  /// It will never emit an error event.
+  /// _It's not a broadcast stream, because cancelling a subscription
+  /// on the stream, and then quickly listening again, may cause the
+  /// same event (the current value) to be seen twice._
+  abstract final Stream<T> updates;
+
+  /// Creates an observable value with the given initial value.
+  factory ObservableValue(T initialValue) = _ObservableValue<T>;
+
+  // Possible method to add:
+  //
+  // /// Prevents further updates to [value].
+  // ///
+  // /// The [value] can no longer be set.
+  // ///
+  // /// The [updates] stream will close current listeners,
+  // /// and close new listeners after sending them the current value.
+  // void freeze();
+}
+
+/// Default implementation of [ObservableValue].
+class _ObservableValue<T> implements ObservableValue<T> {
+  /// The stored value.
+  T _value;
+
+  /// Listeners waiting for updates.
+  final List<MultiStreamController<T>> _listeners = [];
+
+  /// The update stream, created when the first listener is added.
+  Stream<T>? _stream;
+
+  _ObservableValue(this._value);
+
+  @override
+  T get value => _value;
+  @override
+  set value(T value) {
+    _value = value;
+    _notify(value);
+  }
+
+  @override
+  Stream<T> get updates => _stream ??= Stream<T>.multi((controller) {
+        controller.add(_value);
+        _listeners.add(controller);
+      });
+
+  void _notify(T value) {
+    // Sends event to active listeners, and compacts listener list
+    // to only contain listeners which are still active.
+
+    // Since `add` is asynchronous, this loop will not call any user code.
+    int activeCount = 0;
+    for (var i = 0; i < _listeners.length; i++) {
+      var controller = _listeners[i];
+      if (controller.hasListener) {
+        controller.add(value); // Async event, so no synchronous side effects.
+        if (activeCount < i) {
+          _listeners[activeCount] = controller;
+        }
+        activeCount++;
+      }
+    }
+    if (activeCount < _listeners.length) _listeners.length = activeCount;
+  }
+}
--- a/test/observable_value_test.dart
+++ b/test/observable_value_test.dart
@@ -1,0 +1,61 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:async/async.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test("Sends initial value", () async {
+    var o = ObservableValue<int>(1);
+    expect(await o.updates.first, 1);
+  });
+
+  test("Sends updated value at time of listen", () async {
+    var o = ObservableValue<int>(1);
+    o.value = 2;
+    expect(await o.updates.first, 2);
+  });
+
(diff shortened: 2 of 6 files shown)


lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+///   // ...
+///
+///   Stream<bool> get visibilityChanges => _isVisible.updates;
+///
+///   void toggleVisibility() {
+///     _isVisible.value ^= true; // XOR with true toggles value.
+///   }
+/// }
+/// ```
+abstract class ObservableValue<T> {
```
lrhn, line 35: I'm not totally happy with the name.
It's valid, but I fear "Observable" might be overused and it might be mistaken for meaning something specific to another use of the word.

Alternatives welcome!

Maybe `ValueNotifier` (focuses on the notification, not the value, but the notification is what distinguishes it from a normal variable), `StateStream` (too vague), `ValuePublisher` (better, but fear it might be confusing to new readers - Nate, wdyt?).

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+/// A single variable with a stream emitting the variable's changing values.
+///
+/// Contains a single [value] which can be read and written as any other
+/// variable.
+///
+/// Whenever the value is set, the new value is emitted on the
+/// [updates] stream.
+/// Further, any new listener on the [updates] stream is initially sent the
+/// current [value].
+///
```
macjohnny, line 16: maybe we should point out that this has a similar behavior as the [BehaviorSubject](https://pub.dev/documentation/rxdart/latest/rx/BehaviorSubject-class.html) in the RxDart / ReactiveX world.
```suggestion
/// This this a behavior similar to the
/// [BehaviorSubject](https://pub.dev/documentation/rxdart/latest/rx/BehaviorSubject-class.html)
/// in RxDart / ReactiveX.
///
```

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+/// Consider using immutable types for [value] since listeners will not
+/// be notified if the internal state of [value] changes, only if
+/// the [value] variable is written to.
+///
```
macjohnny, line 20:
```suggestion
/// This is particularly useful for a robust redux-like state management approach.
/// 
```

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+  /// of [value] as soon as possible after listening.
+  /// Every time [value] changes, they will then receive an event
+  /// with the new value.
+  ///
+  /// This stream can be listened to multiple times.
+  /// It will never emit an error event.
+  /// _It's not a broadcast stream, because cancelling a subscription
+  /// on the stream, and then quickly listening again, may cause the
+  /// same event (the current value) to be seen twice._
+  abstract final Stream<T> updates;
```
macjohnny, line 54: I suggest to rename this to `values` or `valueStream` to emphasize it is not only updates to the values after listening, but we will get the value that was set so far. The fact that we get updates should be clear from the `Stream` type

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+        controller.add(_value);
+        _listeners.add(controller);
+      });
+
+  void _notify(T value) {
+    // Sends event to active listeners, and compacts listener list
+    // to only contain listeners which are still active.
+
+    // Since `add` is asynchronous, this loop will not call any user code.
+    int activeCount = 0;
```
macjohnny, line 102: how about 
```suggestion
    var activeCount = 0;
```
?

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+  @override
+  set value(T value) {
+    _value = value;
+    _notify(value);
+  }
+
+  @override
+  Stream<T> get updates => _stream ??= Stream<T>.multi((controller) {
+        controller.add(_value);
+        _listeners.add(controller);
```
macjohnny, line 94: I think we should remove the controller from `_listeners` when the subscriber stops listening to avoid a memory leak. Otherwise a controller is kept in memory until a new value is emitted (which might never happen)
```suggestion
        _listeners.add(controller);
        controller.onCancel = () {
          _listeners.remove(controller);
        };
```

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+
+  void _notify(T value) {
+    // Sends event to active listeners, and compacts listener list
+    // to only contain listeners which are still active.
+
+    // Since `add` is asynchronous, this loop will not call any user code.
+    int activeCount = 0;
+    for (var i = 0; i < _listeners.length; i++) {
+      var controller = _listeners[i];
+      if (controller.hasListener) {
```
macjohnny, line 105: nit: for better readability, I suggest to do the inactive listener cleanup in a separate loop.

test/observable_value_test.dart
```diff
@@ -0,0 +1,61 @@
...
+    var sub = o.updates.listen(null);
+    sub.onData(expectAsync1(count: 2, (v) async {
+      expect(v, expectation);
+      if (v == 1) expectation = 2;
+      if (v == 2) sub.cancel();
+      expect(await o.updates.first, 2); // Same event again.
+    }));
+    o.value = 2;
+  });
+}
```
macjohnny, line 61: I think we should also add a test for the value getter

lib/src/observable_value.dart
macjohnny: I would vote for `ValuePublisher`

natebosch:
Overall I'm worried about putting this interface into `package:async` from the start. I'd feel better experimenting in some other place - either internally or in a separate package - so that we can iterate on the API. That was ultimately the conclusion of https://github.com/dart-lang/async/issues/31, but then we never made the push to land something. If we have more motivation now I'm in favor of continuing the exploration.

lib/src/observable_value.dart
lrhn: That would definitely need a link. Until you mentioned it, I had never heard of "redux state mangement".

lib/src/observable_value.dart
lrhn: That would be `values`. No need to put `Stream` on the name of something with type `Stream`.

lib/src/observable_value.dart
lrhn: My fingers write `int` without me even trying. I have to actively avoid it.
(Good think it's not wrong.)

lib/src/observable_value.dart
lrhn: Possible. If there are no inactive listeners, we don't have to call it.
If we do, it's still linear. It's not as efficient, but probably still reasonable.

lib/src/observable_value.dart
macjohnny: > why we only support that one thing

Here we only mention a particularly helpful use case, but in no way we want to suggest that we don't support other use cases.

A nice link might be https://ngrx.io/guide/store/why

lrhn: I meant: Why do we not support the rest of the reactive operations from Rx? Why just the "BehaviorSubject"?

It's a nice primitive, but RxDart already has it.
So, why is `dart:async` providing this particular building block by itself, and not any of the other related ones?

lrhn: Not sure I'm good with writing something this specific. If you already know, you don't need the text saying it, and if you don't, it's *far* too specific to put in front of the average user.

lib/src/observable_value.dart
lrhn: I should check that controllers and subscriptions clean up when they're cancelled, so they won't hang onto unnecessary state.
If they do that (or I make them), then it shouldn't be a problem to retain a controller which has no live listeners or callbacks any more. That's probably a better solution than doing eager clean-up

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+  ///
+  /// The [updates] stream emits the current value when listened to,
+  /// and the new value when `value` is set to a new value.
+  abstract T value;
+
+  /// A stream emitting the current value of [value].
+  ///
+  /// New listeners will receive an event containing the current value
+  /// of [value] as soon as possible after listening.
+  /// Every time [value] changes, they will then receive an event
```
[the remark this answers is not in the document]
lrhn: I asked @macjohnny about that, and decided to publish on all set operations.

You can always choose to use `.distinct` on the stream, using whichever equality is more appropriate.

Or create a subclass with
```dart
void set value(T value) {
  if (!identical(value, this.value)) super.value = value;
}
```
Or just checking before setting, if you don't want to *send* equal values.
All the options are open.

Baking it into the class means either taking a stand on what "distinct" means or adding the overhead of a user-supplied `equals` function to call.

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+  ///
+  /// This stream can be listened to multiple times.
+  /// It will never emit an error event.
+  /// _It's not a broadcast stream, because cancelling a subscription
+  /// on the stream, and then quickly listening again, may cause the
+  /// same event (the current value) to be seen twice._
+  abstract final Stream<T> updates;
+
+  /// Creates an observable value with the given initial value.
+  factory ObservableValue(T initialValue) = _ObservableValue<T>;
```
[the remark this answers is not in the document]
lrhn: That's again trying to build complex functionality into a simple class.

If you want to set the values from a stream, you can do `stream.forEach((value) => _publisher.value = value);`.
You still need to supply an initial value.

Or make a different class implementing the same interface. If it has the same interface. (Is there a value setter?)


The more general question would be: Which problem (or how many problems) are we trying to solve?
Rewriting all of RxDart is not a goal. How much is enough?

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+/// }
+/// ```
+abstract class ObservableValue<T> {
+  /// The current value.
+  ///
+  /// The [updates] stream emits the current value when listened to,
+  /// and the new value when `value` is set to a new value.
+  abstract T value;
+
+  /// A stream emitting the current value of [value].
```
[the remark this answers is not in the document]
lrhn: Rephrased.

lib/src/observable_value.dart
```diff
@@ -0,0 +1,115 @@
...
+///     _isVisible.value ^= true; // XOR with true toggles value.
+///   }
+/// }
+/// ```
+abstract class ObservableValue<T> {
+  /// The current value.
+  ///
+  /// The [updates] stream emits the current value when listened to,
+  /// and the new value when `value` is set to a new value.
+  abstract T value;
```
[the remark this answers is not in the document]
natebosch: > So, again, what problem should we solve. And why should we solve it, if RxDart already exists?

Experimenting with an API internally without publishing could help us identify the specific problems we think this solves. I'd hope we could find at least a few use cases before we decide to publish.

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

Issue #229, opened 2023-01-25
Bad State error while trying to reject a StreamQueueTransaction


natebosch:
```
  Bad state: The FutureGroup is closed.
  package:async/src/future_group.dart 70:18      FutureGroup.add
  package:async/src/stream_splitter.dart 130:37  StreamSplitter._cancelSubscription
  package:async/src/stream_splitter.dart 176:7   StreamSplitter._onCancel
  package:async/src/stream_splitter.dart 80:33   StreamSplitter.split.<fn>
  dart:async                                     _BufferingStreamSubscription.cancel
  package:async/src/stream_queue.dart 492:33     StreamQueue._cancel
  package:async/src/stream_queue.dart 617:13     StreamQueueTransaction._done
  package:async/src/stream_queue.dart 609:5      StreamQueueTransaction.reject
```

Getting in to this state requires some particular interactions with the transaction - one set of interactions I've found that causes it:
- Create a transaction
- Create a new queue from the transaction
- Create a new transaction from the new queue
- Create a third queue from the new transaction, and consume the entire stream
- Reject the new transaction (where we had consumed the stream)
- Create a new queue from the original transaction. Don't consume the stream
- Try to reject the original transaction

I'm not sure if the bug is in `StreamQueue`, or in `StreamSplitter`.

PR #230, opened 2023-01-25
Regression test for error rejecting transaction

natebosch:
Towards #229

Add a skipped test demonstrating the sequence of interaction with a
StreamQueueTransaction that leads to a `StateError` during the call to
`transaction.reject()`.


natebosch:
@lrhn - I think its worth landing the skipped test for now while we work on a fix.

lrhn (approved):

PR #231, opened 2023-01-26
blast_repo fixes

devoncarew:
auto-publish

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,3 +92,8 @@ computations.
 [StreamZip]: https://pub.dev/documentation/async/latest/async/StreamZip-class.html
 [SubscriptionStream]: https://pub.dev/documentation/async/latest/async/SubscriptionStream-class.html
 [typedStreamTransformer]: https://pub.dev/documentation/async/latest/async/typedStreamTransformer.html
+
+## Publishing automation
+
+For information about our publishing automation and release process, see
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -92,3 +92,8 @@ computations.
 [StreamZip]: https://pub.dev/documentation/async/latest/async/StreamZip-class.html
 [SubscriptionStream]: https://pub.dev/documentation/async/latest/async/SubscriptionStream-class.html
 [typedStreamTransformer]: https://pub.dev/documentation/async/latest/async/typedStreamTransformer.html
+
+## Publishing automation
```
natebosch, line 96: WDYT about adding a CONTRIBUTING.md file with this information? It isn't very relevant to consumers of this package on pub.

natebosch (approved):

README.md
devoncarew: Yeah, makes sense. I mostly want to remove any uncertainty about how to contribute to the repo - the form the PR should be in. If we don't want a brief blurb in the readme, a combo of a section in the contributing doc, and having a link written into the publish bot's PR comments should work.

devoncarew: I may land this as is and figure out the right balance between the readme and the contributing docs asynchronously.

PR #232, opened 2023-01-26
Simplify implementation of `StreamQueue`.

lrhn:
Only change in behavior is that all event seen by the original queue while the transaction request is active, are forwarded to the transaction immediately, and will still be visible after the transaction commits or rejects.

The existing behavior was based on cancelling a stream subscription and which events were visible depended on which commands had been used to request those events already.

### DIFF
--- a/lib/src/stream_queue.dart
+++ b/lib/src/stream_queue.dart
@@ -5,12 +5,9 @@
 import 'dart:async';
 import 'dart:collection';
 
-import 'package:collection/collection.dart';
-
 import 'cancelable_operation.dart';
 import 'result/result.dart';
 import 'stream_completer.dart';
-import 'stream_splitter.dart';
 import 'subscription_stream.dart';
 
 /// An asynchronous pull-based interface for accessing stream events.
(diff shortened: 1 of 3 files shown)


lrhn:
WDYT?

lib/src/stream_queue.dart
```diff
@@ -543,28 +600,47 @@ class StreamQueueTransaction<T> {
...
   ///
   /// This copy starts at the parent queue's position when
   /// [StreamQueue.startTransaction] was called. Its position can be committed
   /// to the parent queue using [commit].
   StreamQueue<T> newQueue() {
-    var queue = StreamQueue(_splitter.split());
+    if (!_active) {
+      return StreamQueue._().._flags = StreamQueue._flagDone;
+    }
+    var queue = StreamQueue<T>._init(_parent._eventQueue, _parent._isDone);
```
lrhn, line 643: An alternative would be to make the transaction `StreamQueue` pull its parent's events one at a time, when it has a request enqueued. That way it won't copy events that it hasn't asked for, and then we can cut the link when cancelling.
I think that would bring us close to the current behavior. It's more complicated, but might avoid copying events that aren't needed yet. 

Probably needs an asynchronous delay as well, so it'll have to start a microtask to request events.

natebosch:
This does fix https://github.com/dart-lang/async/issues/229

natebosch:
This causes new failures, both for the existing stream matchers in `package:test_api`, and for the new copies in `package:checks` which use transactions in similar ways.

The easiest way to see the failure for now is to grab the test repo and override the async dep to your local copy or this git branch and run the test `test/frontend/stream_matcher_test.dart`.

```sh
cd pkgs/test_api
```

```sh
echo "  async:
    git:
      url: git://github.com/dart-lang/async
      ref: stream-queue-simp" >> pubspec.yaml
```

```sh
dart test test/frontend/stream_matcher_test.dart
```

Alternatively, grab the branch from https://github.com/dart-lang/test/pull/1868 and use `test/extensions/async_test.dart` from `checks`

```sh
cd pkgs/checks
```

```sh
echo "dependency_overrides:
  async:
    git:
      url: git://github.com/dart-lang/async
      ref: stream-queue-simp" > pubspec_overrides.yaml
```

```sh
dart test test/extensions/async_test.dart
```

I'll try to find some time tomorrow to prune down more narrow reproduction cases, but I may not be able to get to it until next week.

lib/src/stream_queue.dart
natebosch: Pulling does sound like a good approach at a high level, but I haven't thought through any details.

> Probably needs an asynchronous delay as well, so it'll have to start a microtask to request events.

Why is that? When it pulls an even it should get it asynchronously already right?

lrhn:
Seems cancellation is something I need to look closer at. So far it's been happening indirectly by cancelling a stream subscription. It might make sense to build it into the code directly, so each `Request` object has a `cancel` method that can make it shut down properly.

Will try that.

lib/src/stream_queue.dart
lrhn: If adding a request would start pulling events from the parent queue synchronously, until the request is completed, then the request will be able to complete synchronously.
That might not be a problem, if it just completes an asynchronous completer, nobody will know.
I'm just not absolutely sure that everything that happens is asynchronous.

mosuem:
Closing this as it is stale and needs more background work. Feel free to reopen in case you still want to land it!

PR #235, opened 2023-02-08
Add `StreamExtensions.buffer()`

nex3:
The immediate motivation for this is to buffer stderr from a subprocess until we learn whether the subprocess completed successfully or not. It's important to consume the output eagerly so the process doesn't deadlock after saturating the OS and/or `dart:io` buffers, but we don't want to print it at all if the process completes successfully.

I think it's generally useful to have an explicit way of doing this, since otherwise users may naively try to write `Stream.pipe(StreamController)` which doesn't actually buffer output if the controller doesn't have a listener.

### DIFF
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -62,6 +62,11 @@ extension StreamExtensions<T> on Stream<T> {
   /// stream is listened to, at which point all buffered events will be emitted
   /// in order, and then further events from this stream will be emitted as they
   /// arrive.
+  ///
+  /// The buffer will retain all events until the returned stream is listened
+  /// to, so if the stream can emit arbitrary amounts of data callers should be
+  /// careful to listen to it eventually or call `stream.listen(null).cancel()`
+  /// to discard it if it's not needed.
   Stream<T> bufferUntilListen() {
     var controller = StreamController<T>(sync: true);
     var subscription = listen(controller.add,
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -64,9 +64,10 @@ extension StreamExtensions<T> on Stream<T> {
   /// arrive.
   ///
   /// The buffer will retain all events until the returned stream is listened
-  /// to, so if the stream can emit arbitrary amounts of data callers should be
-  /// careful to listen to it eventually or call `stream.listen(null).cancel()`
-  /// to discard it if it's not needed.
+  /// to, so if the stream can emit arbitrary amounts of data, callers should be
+  /// careful to listen to the stream eventually or call
+  /// `stream.listen(null).cancel()` to discard the buffered data if it becomes
+  /// clear that the data isn't not needed.
   Stream<T> bufferUntilListen() {
     var controller = StreamController<T>(sync: true);
     var subscription = listen(controller.add,
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -68,7 +68,7 @@ extension StreamExtensions<T> on Stream<T> {
   /// careful to listen to the stream eventually or call
   /// `stream.listen(null).cancel()` to discard the buffered data if it becomes
   /// clear that the data isn't not needed.
-  Stream<T> bufferUntilListen() {
+  Stream<T> listenAndBuffer() {
     var controller = StreamController<T>(sync: true);
     var subscription = listen(controller.add,
         onError: controller.addError, onDone: controller.close);
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
     });
     return completer.future;
   }
+
+  /// Returns a view of this stream that eagerly buffers all events until
+  /// `listen()` is called.
+  Stream<T> buffer() {
+    late StreamSubscription<T> subscription;
+    var controller = StreamController<T>(
+        onPause: () => subscription.pause(),
+        onResume: () => subscription.resume(),
+        onCancel: () => subscription.cancel());
+    subscription = listen(controller.add,
+        onError: controller.addError, onDone: controller.close);
+    return controller.stream;
+  }
 }
--- a/lib/src/stream_extensions.dart
+++ b/lib/src/stream_extensions.dart
@@ -55,16 +55,21 @@ extension StreamExtensions<T> on Stream<T> {
     return completer.future;
   }
 
-  /// Returns a view of this stream that eagerly buffers all events until
-  /// `listen()` is called.
-  Stream<T> buffer() {
-    late StreamSubscription<T> subscription;
-    var controller = StreamController<T>(
-        onPause: () => subscription.pause(),
-        onResume: () => subscription.resume(),
-        onCancel: () => subscription.cancel());
-    subscription = listen(controller.add,
+  /// Eagerly listens to this stream and buffers events until needed.
+  ///
+  /// The returned stream will emit the same events as this stream, starting
+  /// from when this method is called. The events are delayed until the returned
+  /// stream is listened to, at which point all buffered events will be emitted
+  /// in order, and then further events from this stream will be emitted as they
+  /// arrive.
+  Stream<T> bufferUntilListen() {
+    var controller = StreamController<T>(sync: true);
+    var subscription = listen(controller.add,
         onError: controller.addError, onDone: controller.close);
+    controller
+      ..onPause = subscription.pause
+      ..onResume = subscription.resume
+      ..onCancel = subscription.cancel;
     return controller.stream;
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 * Add `CancelableOperation.fromValue`.
 
-* Add `StreamExtensions.buffer`, which buffers events from a stream before it
-  has a listener.
+* Add `StreamExtensions.listenAndBuffer`, which buffers events from a stream
+  before it has a listener.
 
 ## 2.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 2.11.0
 
 * Add `CancelableOperation.fromValue`.
-
 * Add `StreamExtensions.listenAndBuffer`, which buffers events from a stream
   before it has a listener.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
-## 2.11.0-dev
+## 2.11.0
 
 * Add `CancelableOperation.fromValue`.
+
+* Add `StreamExtensions.buffer`, which buffers events from a stream before it
+  has a listener.
 
 ## 2.10.0
 
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: async
-version: 2.11.0-dev
+version: 2.11.0
 description: Utility functions and classes related to the 'dart:async' library.
 repository: https://github.com/dart-lang/async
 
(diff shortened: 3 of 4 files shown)


lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
     });
     return completer.future;
   }
+
+  /// Returns a view of this stream that eagerly buffers all events until
+  /// `listen()` is called.
+  Stream<T> buffer() {
```
natebosch, line 60: This name conflicts with `buffer` from `stream_transform`.

https://pub.dev/documentation/stream_transform/latest/stream_transform/RateLimit/buffer.html

WDYT about `asSingleSubscriptionStream` to contrast with `asBroadcastStream`?


Would it be sensible to also have

```dart
if (!isBroadcast) return this;
```

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
...
+    late StreamSubscription<T> subscription;
+    var controller = StreamController<T>(
+        onPause: () => subscription.pause(),
+        onResume: () => subscription.resume(),
+        onCancel: () => subscription.cancel());
+    subscription = listen(controller.add,
```
natebosch, line 66: I think it is safe to pass `cancelOnError: false` here.

If the result is listened with `cancelOnError: false`, we'd need it here too or that parameter won't work.

If the result is listened with `cancelOnError: true`, it should be OK. We'll add multiple errors here, but close the subscription once the result's listener closes on the first error.

natebosch (commented):
@lrhn - do you think this would fit better in `stream_transform` while it's still separate from `async`?

natebosch:
> It's important to consume the output eagerly so the process doesn't deadlock after saturating the OS and/or `dart:io` buffers

We don't get _unlimited_ extra buffering here right? There is still backpressure through pausing the subscription, so is it still possible to fill up an OS buffer?

lib/src/stream_extensions.dart
natebosch: > Would it be sensible to also have

I guess if the goal is specifically to have a listen on streams that might not otherwise ever have one it wouldn't make sense to `return this`.

We could also consider the name `eagerListen` instead of `asSingleSubscriptionStream` to reinforce the intent.

nex3: I think `eagerListen()` or `bufferUntilListen()` would make sense as names.

lib/src/stream_extensions.dart
nex3: I agree—but isn't `cancelOnError: false` the default?

nex3:
> We don't get _unlimited_ extra buffering here right? There is still backpressure through pausing the subscription, so is it still possible to fill up an OS buffer?

That's right, it buffers until the point `listen()` is called and then behaves like a normal stream with normal backpressure. We could potentially add a `bufferWhilePaused` parameter as well, either now or in the future.

lib/src/stream_extensions.dart
natebosch: 🤦 - it is the default.

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
     });
     return completer.future;
   }
+
+  /// Returns a view of this stream that eagerly buffers all events until
+  /// `listen()` is called.
```
lrhn, line 59: I'd drop "view of", that's not a useful description. It's a separate stream, based on this stream.

Maybe:
```dart
/// Eagerly listens to this stream and buffers events until needed.
///
/// The returned stream will emit the same events as this stream,
/// starting from when the `buffer` method is called.
/// The events are delayed until the returned stream is listened to,
/// then the buffered events will be emitted, and then further
/// events from this stream will be emitted as they arrive.
```

lib/src/stream_extensions.dart
lrhn, line 60: Consider adding parameters like `([int? maxBuffer])`, which makes the buffering pause after `maxBuffer` events if set. Maybe `([int? maxBuffer, int Function(T)? weight])`, so you can assign a "size" to each event, e.g., the length of lists, and count that towards the maximum buffering size.

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
...
+  Stream<T> buffer() {
+    late StreamSubscription<T> subscription;
```
lrhn, line 61: Please don't use `late`. Just make it nullable. Or, preferably, reorder things so it's not necessary.

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
...
+    var controller = StreamController<T>(
+        onPause: () => subscription.pause(),
+        onResume: () => subscription.resume(),
+        onCancel: () => subscription.cancel());
```
lrhn, line 65: If `subscription` was already created, we wouldn't need the lambda-wrappings.

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,17 @@ extension StreamExtensions<T> on Stream<T> {
...
+    subscription = listen(controller.add,
+        onError: controller.addError, onDone: controller.close);
+    return controller.stream;
```
lrhn, line 68: Consider avoiding `late` and needing wrappers by setting callbacks after creation:
```dart
  var controller = StreamController<T>.sync();
  var subscription = this.listen(controller.add, controller.addError, controller.close);
  controller
    ..onPause = subscription.pause
    ..onResume = subscription.resume
    ..onCancel = subscription.cancel;
  return controller.stream;
```

With `maxBuffer`/`weight` that becomes:
```dart
  if (maxBuffer != null) RangeError.checkNotNegative(maxBuffer, "maxBuffer");
  var controller = StreamController<T>.sync();
  var subscription = this.listen(null, controller.addError, controller.close); 
  if (maxBuffer != null) {
    var bufferedEvents = 0;
    subscription.onData((T value) {
      controller.add(value);
      bufferedEvents += weight?.call(value) ?? 1;
      if (bufferedEvents >= maxBuffer) subscription.pause();
    });
  } else {
    subscription.onData(controller.add);
  }
  controller.onListen = () {
    subscription.onData(controller.add);
    subscription.resume();
  };
  controller
    ..onPause = subscription.pause
    ..onResume = subscription.resume
    ..onCancel = subscription.cancel;
  return controller.stream;
```

lib/src/stream_extensions.dart
lrhn: Or `eagerlyBuffer()`. 
The `bufferUntilListen` seems most descriptive, `eagerListen` sounds like it's _more_ eager than just doing `listen` now, which isn't possible.

lrhn (approved):
Seems reasonable, implementation can be simplified. Feature can potentially be made more feature-full, but that can always be added later if there is an actual need.

lrhn:
I wouldn't put anything new into `stream_transformers`.

Either we like the features in there, or at least some of them, and we should put those into `async`, or we don't, and we should drop them all.
Time to graduate from being an experiment :)

lib/src/stream_extensions.dart
jakemac53: I like `bufferUntilListen` even if it's a bit verbose.

natebosch: I could get on board with `bufferUntilListen` if it is the consensus choice, but I like that `eagerListen` highlights the focus on the listen that happens now (which I find to be the interesting one), rather than the listen that happens later.

How about `listenAndBuffer()`?

nex3: Let's leave this for later if the need arises—I think the main use of this will be to buffer _all_ events, so I'm not sure having buffer size controls are likely to be valuable enough to add a positional parameter.

lrhn: The `listenAndBuffer` is technically correct. It's a low-level (low abstraction) name because it describes the *implementation*, not a higher-level concept.
Maybe because there is no higher level concept, this really is a shorthand for listening-and-buffering, which will only be used as a low-level primitive of other things. Users shouldn't need to use it directly.

lrhn: I am worried about having a function which can cause unlimited memory bloat. I'd really like to force the user to *consider* the problem.

At least put something in the documentation about needing to listen to the stream. Even it it's only `.listen(null).cancel()`, it's needed to stop the underlying stream.

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,22 @@ extension StreamExtensions<T> on Stream<T> {
...
     return completer.future;
   }
+
+  /// Eagerly listens to this stream and buffers events until needed.
+  ///
+  /// The returned stream will emit the same events as this stream, starting
+  /// from when this method is called. The events are delayed until the returned
+  /// stream is listened to, at which point all buffered events will be emitted
+  /// in order, and then further events from this stream will be emitted as they
+  /// arrive.
```
lrhn, line 64: So, add something like:
```dart
/// 
/// The buffering will retain all events until the returned stream is listened to.
/// Therefore the returned stream *must* be listened to, otherwise the buffer
/// can grow arbitrarily large.
/// If the stream is not needed anyway, it should be discarded by doing, at least,
/// `stream.listen(null).cancel();`.

Comment (nex3) [reply lib/src/stream_extensions.dart:64]: I'm not sure "*must* be listened to" is correct—in a lot of cases, this is going to be consuming output from a process or a file where the size is extrinsically constrained and just calling `.bufferUntilListen()` isn't any riskier than something like `File.readAsString()` or `Process.run()`. I rewrote it a bit to make that clearer.

File: lib/src/stream_extensions.dart (hunk 1 above)
Comment (natebosch) [reply lib/src/stream_extensions.dart]: @lrhn - is that a vote for `listenAndBuffer` or do you still prefer `bufferUntilListen`?

I do think `listenAndBuffer` is the least likely to be misread.

Comment (lrhn) [reply lib/src/stream_extensions.dart]: Let's say it's a vote for `listenAndBuffer`. It's an ugly name, but it's precise and understandable.

I like that `listen` is early, because listening to a stream is a significant action to take. You'd usually not expect a function returning a stream to do *anything* before you listen to that stream.

If the name is missing something, it is what it's behavior is after listening to the returned stream.
But adding more would be too long, so putting that in the documentation is fine.

:+1:

File: lib/src/stream_extensions.dart [hunk 8]
```diff
@@ -54,4 +54,27 @@ extension StreamExtensions<T> on Stream<T> {
...
+  ///
+  /// The buffer will retain all events until the returned stream is listened
+  /// to, so if the stream can emit arbitrary amounts of data callers should be
```
lrhn, line 67: Add comma before "callers".

lib/src/stream_extensions.dart
```diff
@@ -54,4 +54,27 @@ extension StreamExtensions<T> on Stream<T> {
...
+  /// careful to listen to it eventually or call `stream.listen(null).cancel()`
+  /// to discard it if it's not needed.
```
lrhn, line 69: Consider changing some (or all) of he "it"s in this sentence with what they refer to.

```dart
  /// careful to listen to the returned stream eventually, 
  /// or do `returnedStream.listen(null).cancel()` to discard the buffered data 
  /// if it becomes clear that the data isn't needed.
```
(Avoiding "it" almost always improves documentation.)

lib/src/stream_extensions.dart
lrhn: That looks fine.

lrhn (approved):

lib/src/stream_extensions.dart
natebosch: @nex3 - can you rename to `listenAndBuffer`? Also update the name in the CHANGELOG.

nex3: Oops, done!

natebosch (approved):

PR #239, opened 2023-04-28
Synchronously forward chained cancellations

natebosch:
Add a test which chains through multiple `.then` calls and then cancels
the original operation. This test passes with the original code, but
starts failing after fixing a bug in `Completer.complete` in the SDK.
https://dart-review.googlesource.com/c/sdk/+/258929

- Add a list of extra completers to forward cancellation. In places
  where the cancellation had been forwarded through a `whenComplete`,
  add to the list of extra cancellations instead. This allows the Future
  return by the outer cancellation to wait for the other callbacks to
  complete before firing.
- When setting up forwarding on an already canceled completer,
  immediately forward the cancellation.
- When forwarding cancellations, don't repeatedly call `_cancel()` on a
  completer which has already been canceled.
- In `race()`, disable cancellation propagation to prevent a cyclic
  cancellation.

### DIFF
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -269,20 +269,11 @@ class CancelableOperation<T> {
                       error2, identical(error, error2) ? stack : stack2);
                 }
               });
-    final forwardingCancel = onCancel == null
-        ? (() => !completer.isCanceled ? completer._cancel() : null)
-        : () async {
-            if (completer.isCanceled) return;
-            try {
-              await onCancel(completer);
-            } catch (error, stack) {
-              completer.completeError(error, stack);
-            }
-          };
+    final cancelForwarder = _CancelForwarder<R>(completer, onCancel);
     if (_completer.isCanceled) {
-      forwardingCancel();
+      cancelForwarder._forward();
     } else {
-      _completer._extraOnCancel.add(forwardingCancel);
+      _completer._cancelForwarders.add(cancelForwarder);
     }
     return completer.operation;
   }
@@ -354,7 +345,12 @@ class CancelableCompleter<T> {
   /// The callback to call if the operation is canceled.
   final FutureOr<void> Function()? _onCancel;
 
-  final List<Future<void>? Function()> _extraOnCancel = [];
+  /// Additional cancellations to forward during cancel;
+  ///
+  /// When a cancelable operation is chained through `then` or `thenOperation` a
+  /// cancellation on the original operation will synchronously cancel the
+  /// chained operations.
+  final List<_CancelForwarder> _cancelForwarders = [];
 
   /// Whether [complete] or [completeError] may still be called.
   ///
@@ -511,21 +507,51 @@ class CancelableCompleter<T> {
 
     if (_inner != null) {
       _inner = null;
-      Future<Object?> allCancels() async {
-        final onCancel = _onCancel;
-        // Legacy uses may return a value through the `void`.
-        FutureOr<Object?> toReturn;
-        if (onCancel != null) {
-          toReturn = onCancel();
-        }
-        if (_extraOnCancel.isNotEmpty) {
-          await Future.wait(_extraOnCancel.map((c) => c()).whereNotNull());
-        }
-        return toReturn is Future ? await toReturn : toReturn;
-      }
-
-      cancelCompleter.complete(allCancels());
+      cancelCompleter.complete(_invokeCancelCallbacks());
     }
     return cancelCompleter.future;
+  }
+
+  /// Invoke [_onCancel] and all callbacks in [_cancelForwarders].
+  ///
+  /// Returns the same value as [_onCancel]. Legacy uses may return a value
+  /// despite the signature having `void` return.
+  Future<Object?> _invokeCancelCallbacks() async {
+    final FutureOr<Object?> toReturn = _onCancel?.call();
+    final forwards = _cancelForwarders.map(_forward).whereNotNull().toList();
+    if (forwards.isNotEmpty) await Future.wait(forwards);
+    return toReturn is Future ? await toReturn : toReturn;
+  }
+}
+
+class _CancelForwarder<T> {
+  final CancelableCompleter<T> completer;
+  final FutureOr<void> Function(CancelableCompleter<T>)? onCancel;
+  _CancelForwarder(this.completer, this.onCancel);
+  Future<void>? _forward() {
+    if (completer.isCanceled) return null;
+    final onCancel = this.onCancel;
+    if (onCancel == null) {
+      return completer._cancel();
+    }
+    try {
+      final result = onCancel(completer);
+      if (result is Future) {
+        return result.catchError(completer.completeErrorIfPending);
+      }
+    } catch (error, stack) {
+      completer.completeErrorIfPending(error, stack);
+    }
+    return null;
+  }
+}
+
+Future<void>? _forward<T>(_CancelForwarder<T> forwarder) =>
+    forwarder._forward();
+
+extension on CancelableCompleter {
+  void completeErrorIfPending(Object error, StackTrace stackTrace) {
+    if (isCompleted) return;
+    completeError(error, stackTrace);
   }
 }
--- a/lib/src/cancelable_operation.dart
+++ b/lib/src/cancelable_operation.dart
@@ -230,8 +230,7 @@ class CancelableOperation<T> {
       FutureOr<void> Function(CancelableCompleter<R>)? onCancel,
       bool propagateCancel = true}) {
     final completer = CancelableCompleter<R>(
-        onCancel:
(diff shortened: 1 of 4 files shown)


lib/src/cancelable_operation.dart
```diff
@@ -81,7 +81,8 @@ class CancelableOperation<T> {
     // they're not actually cancelled by this.
     Future<void> cancelAll() {
       done = true;
-      return Future.wait(operations.map((operation) => operation.cancel()));
+      return Future.wait(operations.map((operation) =>
+          !operation.isCanceled ? operation.cancel() : Future.value()));
```
lrhn, line 85: Can also use a list literal to not add an element for cancelled operations:
```dart
   return Future.wait([for (var operation in operations)
      if (!operation.isCanceled) operation.cancel()
   ]);
```

lib/src/cancelable_operation.dart
```diff
@@ -262,16 +263,21 @@ class CancelableOperation<T> {
...
+        ? () => !completer.isCanceled ? completer._cancel() : Future.value()
         : () async {
             if (completer.isCanceled) return;
             try {
               await onCancel(completer);
             } catch (error, stack) {
               completer.completeError(error, stack);
             }
-          });
+          };
```
lrhn, line 275: Instead of creating functions here, storing them in `_extraOnCancel`, and calling them at the other end, 
could you just store pairs of `(CancelableCompleter completer, FutureOr<void> Function(CancelableCompleter<R>)? onCancel)`, and then handle all this logic at the consumption point?
(Like, have a `Future<void>? _callExtraOnCancel(CancelableCompleter completer, FutureOr<void> Function(CancelableCompleter<R>)? onCancel) { ... }` which does what the two functions above would given the same values.)

Then the code below can be:
```dart
   if (_completer.isCanceled) {
     Timer(Duration.zero, () {
       _callExtraOnCancel(completer, onCancel);
     });
   } else {
      _completer._extraOnCancel.add((completer, onCancel));
   }
```
(May need a helper class if not 3.0 ready yet.)
It seems those two values are all the closures use anyway.

lib/src/cancelable_operation.dart
```diff
@@ -342,6 +348,8 @@ class CancelableCompleter<T> {
   /// The callback to call if the operation is canceled.
   final FutureOr<void> Function()? _onCancel;
 
+  final List<Future<void> Function()> _extraOnCancel = [];
```
lrhn, line 351: Document this. Maybe:

> ```dart
> /// Extra operations to be executed when the operation is canceled.
> ///
> /// Added when .... something ... propagating cancels ... something.
> ```

lrhn (commented):
I have far too many ideas.

In general, we should avoid unnecessary overhead in library code. If users have performance issues should not come from our code, not even because of a thousand tiny microtask cuts.

lib/src/cancelable_operation.dart
natebosch: > could you just store pairs of `(CancelableCompleter completer, FutureOr<void> Function(CancelableCompleter<R>)? onCancel)`

Do we have any syntax to express a generic record type?

I don't think this will work as a list of records without being able to match the generics between the completer and onCancel

natebosch: I added a `_CancelForwarder` class to avoid some of these eager closures.

lib/src/cancelable_operation.dart
```diff
@@ -262,16 +263,21 @@ class CancelableOperation<T> {
...
                 }
               });
-    _completer._cancelCompleter?.future.whenComplete(onCancel == null
-        ? completer._cancel
+    final forwardingCancel = onCancel == null
+        ? () => !completer.isCanceled ? completer._cancel() : Future.value()
         : () async {
             if (completer.isCanceled) return;
             try {
               await onCancel(completer);
```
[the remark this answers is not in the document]
natebosch: This is now avoiding `async` in the implementation of `_CancelForwarder._forward()`.

lib/src/cancelable_operation.dart
```diff
@@ -497,8 +505,20 @@ class CancelableCompleter<T> {
...
-      cancelCompleter.complete(onCancel == null ? null : Future.sync(onCancel));
+      Future<Object?> allCancels() async {
+        final onCancel = _onCancel;
+        // Legacy uses may return a value through the `void`.
+        FutureOr<Object?> toReturn;
+        if (onCancel != null) {
+          toReturn = onCancel();
+        }
+        if (_extraOnCancel.isNotEmpty) {
+          await Future.wait(_extraOnCancel.map((c) => c()));
```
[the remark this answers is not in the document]
natebosch: This now only has the extra `await` when `toReturn` is a Future. `return toReturn is Future ? await toReturn : toReturn;`

lib/src/cancelable_operation.dart
```diff
@@ -497,8 +505,20 @@ class CancelableCompleter<T> {
...
+        }
+        return await toReturn;
+      }
```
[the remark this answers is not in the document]
natebosch: I moved the implementation to an `async` member on `CancelableCompleter` so there isn't a closure involved.

lib/src/cancelable_operation.dart
natebosch: Done.

lib/src/cancelable_operation.dart
lrhn: No generic record types. I'd probably just go with `(CancelableCompleter<Object?> completer, FutureOr<void> Function(CancelableCompleter<Object?>)?`, the value type `R` isn't needed when all we do is cancel.

lrhn (approved):
Only the potentially unhandled future needs to be fixed. Rest is more ideas :)

lib/src/cancelable_operation.dart
```diff
@@ -497,9 +507,50 @@ class CancelableCompleter<T> {
...
     }
     return cancelCompleter.future;
   }
+
+  /// Invoke [_onCancel] and forward to other completers in [_cancelForwarders].
+  ///
+  /// Returns the same value as [_onCancel]. Legacy uses may return a value
+  /// despite the signature having `void` return.
+  Future<Object?> _invokeCancelCallbacks() async {
+    final FutureOr<Object?> toReturn = _onCancel?.call();
```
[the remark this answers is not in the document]
natebosch: Updated to use a single `await`. It's another nested ternary conditional, but the formatting looks better in this case 😉

lib/src/cancelable_operation.dart
```diff
@@ -337,11 +340,18 @@ class CancelableCompleter<T> {
...
 
   /// The callback to call if the operation is canceled.
   final FutureOr<void> Function()? _onCancel;
 
+  /// Additional cancellations to forward during cancel.
+  ///
+  /// When a cancelable operation is chained through `then` or `thenOperation` a
+  /// cancellation on the original operation will synchronously cancel the
+  /// chained operations.
+  final List<_CancelForwarder> _cancelForwarders = [];
```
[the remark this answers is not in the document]
natebosch: Done.

lib/src/cancelable_operation.dart
```diff
@@ -497,9 +507,50 @@ class CancelableCompleter<T> {
...
+      if (result is Future) {
+        return result.catchError(completer.completeErrorIfPending);
+      }
+    } catch (error, stack) {
+      completer.completeErrorIfPending(error, stack);
+    }
+    return null;
+  }
+}
+
```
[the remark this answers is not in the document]
natebosch: Done.

lib/src/cancelable_operation.dart
```diff
@@ -497,9 +507,50 @@ class CancelableCompleter<T> {
...
+Future<void>? _forward<T>(_CancelForwarder<T> forwarder) =>
```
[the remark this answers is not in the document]
natebosch: Done.

lib/src/cancelable_operation.dart
```diff
@@ -497,9 +507,59 @@ class CancelableCompleter<T> {
...
+    final cancelFutures = <Future<Object?>>[
+      if (isFuture) toReturn,
+      ...?_cancelForwarders?.map(_forward).whereNotNull()
+    ];
+    final results = (isFuture && cancelFutures.length == 1)
+        ? [await toReturn]
+        : cancelFutures.isNotEmpty
+            ? await Future.wait(cancelFutures)
+            : const [];
+    return isFuture ? results.first : toReturn;
```
lrhn, line 531: LGTM. But just because I can't not optimize, consider:

```dart
  if (cancelFutures.isEmpty) return toReturn;
  if (cancelFutures.length == 1) {
     var result = await cancelFutures[0];
     return isFuture ? result : toReturn;
  }
  var results = await Future.wait(cancelFutures);
  return isFuture ? results[0] : toReturn;
```

(Probably not worth it.)

PR #249, opened 2023-08-16
Added Condition and boundedForEach

jonasfj:
Adding:
 * `Notifier` with `wait` that'll be resolved when `notify()` is called. This is a primitive for signaling.
 * `Stream.parallelForEach(maxParallel, eachFn, onError)` for calling `eachFn` in parallel for each item in the stream.

It's possible that `onError` really should be called `ignoreError` and only take `Object e` or even `Exception e`. Since:
 * You shouldn't use the stack trace to decide if you are ignoring an error.
 * You should never ignore an error, only exceptions.

Or maybe we should call it `continueIf`, because it really makes sense to write: 
`continueIf: (Exception e) => e is TimeoutException` and things like that.

Having an `onError` method where you can choose to rethrow or not is really bad. Because `Error.throwWithStackTrace` is annoying to call, so people are likely to loose their stack traces.


lib/src/condition_variable.dart
```diff
@@ -0,0 +1,65 @@
...
+  /// new future. This future will also be unresolved, until [notify] is called.
+  ///
+  /// The [Future] return from this condition variable will never throw.
+  Future<void> get wait {
+    if (_completer.isCompleted) {
+      _completer = Completer();
+    }
+    return _completer.future;
+  }
+}
```
lrhn, line 65: This is equivalent to a broadcast `Stream<void>` where you can only wait for the next event. An implementation could be:
```dart
class Notifier {
  final _control = StreamController<void>.broadcast();
  void notify() { _control.add(null); }
  Future<void> get wait => _control.stream.first;
}
```
(Probably more expensive, though.)

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,71 @@ extension StreamExtensions<T> on Stream<T> {
       ..onCancel = subscription.cancel;
     return controller.stream;
   }
+
+  /// Invoke [each] for each item in this stream, and wait for the [Future]
+  /// returned by [each] to be resolved. Running no more than [concurrency]
+  /// number of [each] calls at the same time.
```
lrhn, line 86: Start with single-line first paragraph.

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,71 @@ extension StreamExtensions<T> on Stream<T> {
...
+    StackTrace? firstStackTrace;
+
+    var running = 0;
+    final wakeUp = ConditionVariable();
+
+    try {
+      var doBreak = false;
+      await for (final item in this) {
+        running += 1;
+        scheduleMicrotask(() async {
```
lrhn, line 111: Don't think the `scheduleMicrotask` is needed. 
The code is running inside an `await for`, so we're definitely past the initial synchronous part of the `async` functions, and already running inside an event callback.

You can still run the code inside a separate async function, as `() async { ... }();`, without using `scheduleMicrotask`. (Might need to use `unawaited`, but passing an async function to `scheduleMicrotask` is already a code-smell, since it doesn't look at the future anyway.)

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,71 @@ extension StreamExtensions<T> on Stream<T> {
...
+  /// subsequent calls
+  ///
+  /// will
+  /// call [onError], if [onError] throws (it does)
+  /// continue subsequent [each] calls, ignore additional errors and throw the
+  /// first error encountered.
+  Future<void> boundedForEach(
+    int concurrency,
+    FutureOr<void> Function(T item) each, {
+    FutureOr<void> Function(Object e, StackTrace? st) onError = _throwOnError,
```
lrhn, line 99: `_throwOnError` -> `_Future<void>.error`. Just use the constructor tear-off. It's constant.

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,71 @@ extension StreamExtensions<T> on Stream<T> {
...
+  /// This function will wait for the futures returned by [each] to be resolved
+  /// before completing. If any [each] invocation throws, [boundedForEach] stop
+  /// subsequent calls
+  ///
+  /// will
+  /// call [onError], if [onError] throws (it does)
+  /// continue subsequent [each] calls, ignore additional errors and throw the
+  /// first error encountered.
+  Future<void> boundedForEach(
+    int concurrency,
```
lrhn, line 97: Maybe `maxParallel`

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,71 @@ extension StreamExtensions<T> on Stream<T> {
...
+      while (running >= concurrency) {
+        await wakeUp.wait;
+      }
+    }
+
+    final firstError_ = firstError;
+    if (firstError_ != null) {
+      return Future.error(firstError_, firstStackTrace);
+    }
+  }
```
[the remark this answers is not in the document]
jonasfj: How do you rethrow an exception while preserving the existing stacktrace?

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,71 @@ extension StreamExtensions<T> on Stream<T> {
...
+  ///
+  /// This function will wait for the futures returned by [each] to be resolved
+  /// before completing. If any [each] invocation throws, [boundedForEach] stop
+  /// subsequent calls
+  ///
+  /// will
+  /// call [onError], if [onError] throws (it does)
+  /// continue subsequent [each] calls, ignore additional errors and throw the
+  /// first error encountered.
+  Future<void> boundedForEach(
```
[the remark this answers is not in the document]
jonasfj: How about just `parallelForEach` or `concurrentForEach`

I'm leaning away from using the term parallel, because there is no parallelism here, just concurrency.

lib/src/stream_extensions.dart
jonasfj: `maxConcurrency`?

lib/src/stream_extensions.dart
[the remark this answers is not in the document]
jonasfj: Today, I learned about: [`Error.throwWithStackTrace`](https://api.dart.dev/stable/3.1.0/dart-core/Error/throwWithStackTrace.html)

lib/src/stream_extensions.dart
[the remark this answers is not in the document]
jonasfj: `parallelForEach` it is!

lib/src/stream_extensions.dart
```diff
@@ -78,4 +80,90 @@ extension StreamExtensions<T> on Stream<T> {
...
+              doBreak = true;
+              if (firstError == null) {
+                firstError = e;
+                firstStackTrace = st;
+              }
+            }
+          } finally {
+            // When [each] is done, we decrement [running] and notify
+            running -= 1;
+            itemDone.notify();
```
lrhn, line 144: You don't really need `notify` for this.
```dart
Future<void> parallelForEach(
    int maxParallel,
    FutureOr<void> Function(T item) each) {
  var pending = 0;
  var done = false;
  var result = Completer<void>.sync();
  (Object, StackTrace)? firstError;
  void complete() {
    assert(pending == 0);
    assert(done);
    if (firstError case (var error, var stack)) {
      result.completeError(error, stack);
    } else {
      result.complete(null);
    }
 }
  var subscription = stream.listen(null, onDone: () {
    done = true;
    if (pending == 0) {
      complete();
    }
  });
  subscription
    ..onError((Object error, StackTrace stack) {
      subscription.cancel().ignore();
      done = true;
      firstError = (error, stack); // Takes precedence over user errors.
      if (pending == 0) complete();
    })
    ..onData((T value) {
        try {
          var computation = each(value);
          if (computation is Future) {
            if (++pending >= maxParallel) subscription.pause();
            computation.then((_) {
              if (--pending == 0 && done) complete();
              subscription.resume();
            }, onError: (Object error, StackTrace stack) {
              subscription.cancel().ignore();
              done = true;
              firstError ??= (error, stack);
              if (--pending == 0) complete();
           });
        } catch (error, stack) {
          subscription.cancel().ignore();
          done = true;
          firstError ??= (error, stack);
          if (pending == 0) complete();
        }
    });
  return result.future;
}
```

mosuem:

Issue #257, opened 2023-12-05
Async Cache is caching exceptions


akmalviya03:
### Steps to reproduce

To reproduce the same issue you can refer the code sample given below.

### Expected results

It should only cache successful result.

### Actual results

It should not cache exception.

### Code sample

<details><summary>Temporary Cache Code sample</summary>

```dart
class TemporaryCache extends StatefulWidget {
  const TemporaryCache({super.key});

  @override
  State<TemporaryCache> createState() => _TemporaryCacheState();
}

class _TemporaryCacheState extends State<TemporaryCache> {
  late CancelableOperation<String> cancelableOperation;

  Future<String> _futureToRun() async {
    await Future.delayed(const Duration(seconds: 5));
    return Future.value("Temporary Cache");
  }

  Future<String> getData() async {
    return CacheManager().temporaryCache.fetch(() async {
      String? value =  await cancelableOperation.valueOrCancellation();
      if(value != null){
        return value;
      }
      throw Exception('Future Cancelled');
    });
  }

  @override
  void initState() {
    super.initState();
    cancelableOperation =
        CancelableOperation.fromFuture(_futureToRun(), onCancel: () {
          debugPrint('We have successfully cancelled future');
        });
  }

  @override
  void deactivate() async {
    super.deactivate();
    if(!cancelableOperation.isCanceled && !cancelableOperation.isCompleted){
      await cancelableOperation.cancel();
    }
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: FutureBuilder<String>(
            future: getData(),
            builder: (context, snapshot) {
              if(snapshot.hasError){
                Object? error = snapshot.error;
                if(error is Exception){
                  return  Text('$error');
                }else{
                  return const Text('Unknown Error');
                }
              }
              else if (snapshot.connectionState == ConnectionState.done && snapshot.hasData) {
                String? data = snapshot.data;
                if (data != null) {
                  return Text(data);
                }
                return const Text('Data is Null');
              } else {
                return const CircularProgressIndicator();
              }
            }),
      ),
      bottomNavigationBar: SafeArea(
        child: Row(
          mainAxisAlignment: MainAxisAlignment.spaceEvenly,
          mainAxisSize: MainAxisSize.max,
          children: [
            FloatingActionButton.extended(
                label: const Text('Invalidate Temporary Cache'),
                onPressed: () {
                  CacheManager().invalidateTemporaryCache();
                }),
[...]

PR #258, opened 2023-12-05
Async cache won't store exceptions

akmalviya03:
Earlier async cache was caching exceptions.
With new changes in `fetch` method of `AsyncCache` class will no longer store exception.
By default `AsyncCache` will store exception, by setting `_canCacheException` to false. It won't store exception.

Please refer #257 for video demo of both issue and solution along with minimal reproducible code.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -42,7 +42,7 @@ class AsyncCache<T> {
   ///If we set this variable to false
   ///On the initial run, if callback returned the [Exception]
   ///Next time, we can reRun the callback for the successful attempt.
-  final bool _canCacheException;
+  final bool _cacheErrors;
 
   /// Fires when the cache should be considered stale.
   Timer? _stale;
@@ -52,9 +52,9 @@ class AsyncCache<T> {
   /// The [duration] starts counting after the Future returned by [fetch]
   /// completes, or after the Stream returned by `fetchStream` emits a done
   /// event.
-  AsyncCache(Duration duration, {bool canCacheException = true})
+  AsyncCache(Duration duration, {bool cacheErrors = true})
       : _duration = duration,
-        _canCacheException = canCacheException;
+        _cacheErrors = cacheErrors;
 
   /// Creates a cache that invalidates after an in-flight request is complete.
   ///
@@ -63,7 +63,7 @@ class AsyncCache<T> {
   /// data is updated frequently but stale data is acceptable.
   AsyncCache.ephemeral()
       : _duration = null,
-        _canCacheException = true;
+        _cacheErrors = true;
 
   /// Returns a cached value from a previous call to [fetch], or runs [callback]
   /// to compute a new one.
@@ -74,7 +74,7 @@ class AsyncCache<T> {
     if (_cachedStreamSplitter != null) {
       throw StateError('Previously used to cache via `fetchStream`');
     }
-    if (_canCacheException) {
+    if (_cacheErrors) {
       return _cachedValueFuture ??= callback()
         ..whenComplete(_startStaleTimer).ignore();
     } else {
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -81,22 +81,13 @@ class AsyncCache<T> {
       return _cachedValueFuture ??= callback()
         ..whenComplete(_startStaleTimer).ignore();
     } else {
-      if (_cachedValueFuture == null) {
-        try {
-          ///First we run the callback then we assign the value received
-          ///from [callback] to the [_cachedValueFuture]
-          var value = await callback();
-          _cachedValueFuture ??= Future.value(value);
-          _startStaleTimer();
-          return _cachedValueFuture!;
-        ///If [callback] generated an exception then we should not cache data
-        ///And propagate exception to the place from where [fetch] is triggered
-        } catch (error) {
-          rethrow;
-        }
-      } else {
-        return _cachedValueFuture!;
-      }
+      return _cachedValueFuture ??= callback().then((value) {
+        _startStaleTimer();
+        return value;
+      }, onError: (Object error, StackTrace stack) {
+        invalidate();
+        throw error;
+      });
     }
   }
 
--- a/lib/src/async_cache.dart
+++ b/lib/src/async_cache.dart
@@ -38,10 +38,10 @@ class AsyncCache<T> {
   /// Cached results of a previous [fetch] call.
   Future<T>? _cachedValueFuture;
 
-  ///Default is set to true
-  ///If we set this variable to false
-  ///On the initial run, if callback returned the [Exception]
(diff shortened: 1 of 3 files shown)


lib/src/async_cache.dart
```diff
@@ -36,6 +38,12 @@ class AsyncCache<T> {
   /// Cached results of a previous [fetch] call.
   Future<T>? _cachedValueFuture;
 
+  ///Default is set to true
+  ///If we set this variable to false
+  ///On the initial run, if callback returned the [Exception]
```
lrhn, line 43: It's not necessarily an `Exception`, it's the same if the future could completes with any error.

test/async_cache_test.dart
```diff
@@ -18,6 +18,14 @@ void main() {
     cache = AsyncCache(const Duration(hours: 1));
   });
 
+  test('should not fetch when callback throws exception', () async {
+    Future<String> asyncFunctionThatThrows() {
+      throw Exception();
+    }
+    var cacheFuture = cache.fetch(asyncFunctionThatThrows);
+    await expectLater(cacheFuture,throwsA(isException));
```
lrhn, line 26: This doesn't test that the cache is cleared on an error.

Consider:
```dart
var throwingFuture = cache.fetch(() async => throw StateError("Test"));
var secondThrowingFuture = cache.fetch(() async => "Unreachable");
// Future *is* cached until future completes.
expect(throwingFuture, same(secondThrowingFuture));

await expectLater(throwingFuture, throwsA(isStateError));

// Cache is cleared immediately after error future completes.
var valueFuture = cache.fetch(() async => "value");
await expectLater(valueFuture, "value");
```

lib/src/async_cache.dart
```diff
@@ -44,14 +52,18 @@ class AsyncCache<T> {
   /// The [duration] starts counting after the Future returned by [fetch]
   /// completes, or after the Stream returned by `fetchStream` emits a done
   /// event.
-  AsyncCache(Duration duration) : _duration = duration;
+  AsyncCache(Duration duration, {bool canCacheException = true})
```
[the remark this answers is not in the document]
akmalviya03: Done

lib/src/async_cache.dart
akmalviya03:
```
  /// Default is set to true
  /// If this variable is set to false,
  /// and callback gets completed with an error,
  /// then the response will not get cached.
```

How about this description?

lib/src/async_cache.dart
```diff
@@ -36,6 +38,12 @@ class AsyncCache<T> {
   /// Cached results of a previous [fetch] call.
   Future<T>? _cachedValueFuture;
 
+  ///Default is set to true
+  ///If we set this variable to false
+  ///On the initial run, if callback returned the [Exception]
+  ///Next time, we can reRun the callback for the successful attempt.
+  final bool _canCacheException;
```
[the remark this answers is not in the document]
akmalviya03: Thank you for correcting me.

lib/src/async_cache.dart
akmalviya03: Updated description with the suggested one.

lib/src/async_cache.dart
```diff
@@ -62,8 +74,27 @@ class AsyncCache<T> {
...
+          ///from [callback] to the [_cachedValueFuture]
+          var value = await callback();
+          _cachedValueFuture ??= Future.value(value);
+          _startStaleTimer();
+          return _cachedValueFuture!;
+        ///If [callback] generated an exception then we should not cache data
+        ///And propagate exception to the place from where [fetch] is triggered
+        } catch (error) {
+          rethrow;
+        }
```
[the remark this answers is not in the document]
akmalviya03: Sorry, I forgot error always propagate to the place from here function is invoked. You are right we don't need this.

lib/src/async_cache.dart
```diff
@@ -62,8 +74,27 @@ class AsyncCache<T> {
...
-      ..whenComplete(_startStaleTimer).ignore();
+    if (_canCacheException) {
+      return _cachedValueFuture ??= callback()
+        ..whenComplete(_startStaleTimer).ignore();
+    } else {
+      if (_cachedValueFuture == null) {
+        try {
+          ///First we run the callback then we assign the value received
+          ///from [callback] to the [_cachedValueFuture]
+          var value = await callback();
```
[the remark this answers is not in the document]
akmalviya03: Rite, we were waiting for the callback to complete then, we were assigning the value to the cache. While the callback is getting executed, we were able to execute callback once again because value inside the cache was not updated.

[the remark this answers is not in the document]
akmalviya03: We cannot use rethrow because it requires a catch clause

test/async_cache_test.dart
akmalviya03: `expect(throwingFuture, same(secondThrowingFuture));`

What are we trying to check here ?

akmalviya03: Getting this error 

```
Expected: same instance as <Instance of 'Future<String>'>
Actual: <Instance of 'Future<String>'>
```

akmalviya03: I think we are getting this error because both functions are getting different memory address. And it is not able to distinguish.

Attaching supporting links [same method]("https://pub.dev/documentation/test/latest/test/same.html")

MDSADABWASIM:
Any update on this @akmalviya03 @lrhn

lrhn (approved):
SGTM.
@natebosch WDYT?

natebosch:
> @natebosch WDYT?

This one doesn't have the same issue with a `Future` field that loses it's expected association like #260

I don't think I have any concerns about landing this one.

They are interesting to consider together though - adding this capability might give stronger motivation for accepting #260

lib/src/async_cache.dart
```diff
@@ -44,14 +53,18 @@ class AsyncCache<T> {
   /// The [duration] starts counting after the Future returned by [fetch]
   /// completes, or after the Stream returned by `fetchStream` emits a done
   /// event.
-  AsyncCache(Duration duration) : _duration = duration;
+  AsyncCache(Duration duration, {bool cacheErrors = true})
```
natebosch, line 56:
```suggestion
  /// If [cacheErrors] is `false` the cache will be invalidated if the [Future]
  /// returned by the callback completes as an error.
  AsyncCache(Duration duration, {bool cacheErrors = true})
```

natebosch (approved):

lib/src/async_cache.dart
akmalviya03: We have given usage of `cacheErrors` above where `_cacheErrors`  is declared

akmalviya03:
@natebosch @lrhn Any update on this ?

mosuem:

PR #267, opened 2024-02-13
Update to latest lints, require Dart 3.2

kevmoo:


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

test/result/result_test.dart
```diff
@@ -263,7 +263,7 @@ void main() {
   test('handle unary', () {
     var result = ErrorResult('error', stack);
     var called = false;
-    result.handle((error) {
+    result.handle((Object? error) {
```
natebosch, line 266: Does this work if the argument is non-nullable?

`error` instances should never be `null`, and the `stackTrace` should always be a `StackTrace`, but I don't recall if some of the APIs require more loosely typed callbacks

natebosch (approved):

test/result/result_test.dart
kevmoo: There were a BUNCH of weird test failures here. It was really annoying.

A agree. 🤷

